PR #7, opened 2022-02-11
VOD transcode task

darkdarkdragon:
VOD transcode task


go.mod
```diff
@@ -4,12 +4,22 @@ go 1.16
...
+	github.com/livepeer/livepeer-data v0.4.12-0.20220211073659-9cd7e6e820e1
+	github.com/livepeer/stream-tester v0.11.4-0.20220210094813-3f56876c3ac8
 	github.com/peterbourgon/ff v1.7.1
 	github.com/rabbitmq/amqp091-go v1.1.0
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	gopkg.in/vansante/go-ffprobe.v2 v2.0.3
 )
+
+// replace (
```
victorges, line 19: Oh good idea. I've been stashing and popping all the time instead of just keeping this commented here lol

task/transcode.go
```diff
@@ -0,0 +1,237 @@
...
+			break
+		}
+		if seg.Err != nil {
+			err = seg.Err
+			glog.Errorf("===> got error %v", err)
+			break
+		}
+		glog.V(model.DEBUG).Infof("Got segment seqNo=%d pts=%s dur=%s data len bytes=%d\n", seg.SeqNo, seg.Pts, seg.Duration, len(seg.Data))
+		started := time.Now()
+		transcoded, err = lapi.PushSegment(stream.ID, seg.SeqNo, seg.Duration, seg.Data)
```
victorges, line 88: Hmm, I thought we would be able to use the `livepeer-transcode` CLI for the transcode task. Do you think there would any fundamental limitations for evolving to that? Does it support only seekable files as input? If so, would a `bytes.Reader` be enough on that case, since we are relying on the buffered file in memory anyway? Or we could also download the file to disk and then send it to the CLI, which would also have a lower footprint 🤔

task/transcode.go
```diff
@@ -0,0 +1,237 @@
...
+		inputPlaybackID = tctx.InputAsset.PlaybackID
+		lapi            = tctx.lapi
+	)
+
+	fullPath := videoFileName(inputPlaybackID)
+	fir, err := tctx.inputOS.ReadData(ctx, fullPath)
+	if err != nil {
+		panic(err)
+	}
+	fileInMem, err := io.ReadAll(fir.Body)
```
victorges, line 45: Would it be possible/easy to buffer this file on disk instead of in memory? We are currently running the `task-runner` with a `4Gi` memory limit, and I think we should ideally not use that much, so it would be nice to lower the footprint as well. 

Are the outputs being buffered as well btw? Could they be "buffered" on disk instead?

victorges (approved):
LGTM! One question regarding using `livepeer-transcode` and possibly buffering the file on disk instead of in memory.

task/transcode.go
```diff
@@ -0,0 +1,237 @@
...
+	var (
+		ctx             = tctx.Context
+		inputPlaybackID = tctx.InputAsset.PlaybackID
+		lapi            = tctx.lapi
+	)
+
+	fullPath := videoFileName(inputPlaybackID)
+	fir, err := tctx.inputOS.ReadData(ctx, fullPath)
+	if err != nil {
+		panic(err)
```
victorges, line 43: Better to just return the errors instead of panic

task/transcode.go
darkdarkdragon: Yes, it's the major point for optimisation.
But I was not sure what is available in our k8s, and even if `os.TempDir` will return writeable dir. So I went with safest option for now.

task/transcode.go
victorges: I think it was mainly 2 points:
 - Avoiding repeating the logic for tnrascoding files both on the `livepeer-transcode` and on the task-runner
 - Use the task-runner as an internal dog fooder of the `livepeer-transcode` tool. Any improvement we made for the task runner would necessarily be gained by the public users of the CLI tool as well

And if we wanted to avoid the inefficiency of spawning a separate process, we could also break up the `main` function of it just like we did for the `analyzer`, so it can be used as an internal dependency of another package by calling some `Run` function that receives all the CLI args as a parameter: 
https://github.com/livepeer/livepeer-data/blob/c87020230260e838b88223ff017bf2e56e5df0c1/cmd/analyzer/analyzer.go#L108 (in the case of the `analyzer`, would also need to move the `parseFlags` outside `Run` and receive the `cli` as an arg instead)

task/transcode.go
victorges: Alright makes sense! Thx for sharing your concerns

task/transcode.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	fileInMem, err := io.ReadAll(fir.Body)
+	fir.Body.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	streamName := fmt.Sprintf("vod_%s", time.Now().Format("2006-01-02T15:04:05Z07:00"))
+	profiles := tctx.Params.Transcode.Profiles
+	stream, err := lapi.CreateStreamEx(streamName, false, nil, profiles...)
```
victorges, line 53: I think we might want to stick to a single `Profile`, at least on the first version, so that we don't need to make the `outputAssetId` an array instead which will be more complex to handle and index in the Postgres queries (we have queries to find the originating task from a given asset, for example). 

I remember that it is possible to index arrays in Postgres JSONs, but it required some different index type which we don't use yet and had some caveats, so we might to leave that for later. I think users will not be looking for multiple transcoding profiles on this API anyway, at least at first. Will probably make sense for some kind of ABR playback of an asset, in which case I'm not sure we will want to break up the result in multiple assets anyway.

Soo... WDYT of keeping the output a single `outputAssetId`, and either changing the task parameters to a single `Profile` or keeping it an array but checking both here and in the API that only 1 profile is present? That way the schema is a little more future-proof but we can make sure only 1 output asset is created right now.

darkdarkdragon: I've change code everywhere to use only one profile

Issue #12, opened 2022-03-14
All new assets get a automatic `{record: true}` transcode


iameli:
Built on top of livepeer/transcode-cli#8

We want all of these uploaded assets to end up with playback URLs, which for now we're going to create with the recording feature.

victorges:
This is done!

PR #16, opened 2022-04-11
task: prepare

figintern:
https://github.com/livepeer/task-runner/issues/12

First pass at `Prepare` task which creates new stream with `{ record: true }` to get recording of stream for `import` and `transcode` tasks

Currently implemented as a helper but can be moved into its own `task`

No `session` is created when setting `record: true` so the recording streamID is used as the `PlaybackRecordingID` for HLS constructing HLS playback.

Testing on staging:
```
# IMPORT
curl --location --request POST 'https://livepeer.monster/api/asset/import' \
--header 'Authorization: Bearer a001ecb9-05d6-4878-a5af-1ffec64ade40' \
--header 'Content-Type: application/json' \
--data-raw '{
    "url":"https://file-examples.com/storage/fef8fbdce362705a7927afd/2017/04/file_example_MP4_1280_10MG.mp4",
    "name":"tonytest1"
}'

# TRANSCODE
curl --location --request POST 'https://livepeer.monster/api/asset/transcode' \
--header 'Authorization: Bearer a001ecb9-05d6-4878-a5af-1ffec64ade40' \
--header 'Content-Type: application/json' \
--data-raw '{
    "assetId":"9a6fc70f-e1b5-4e8b-837c-a6952fe1ad5e",
   "name": "tonytranscode156",
   "profile": { "name": "480p0", "fps": 30, "bitrate": 100000, "width": 480, "height": 270 }
}'

# LIST ASSETS
~/livepeer/task-runner % curl --location --request GET 'https://livepeer.monster/api/asset' \
--header 'Authorization: Bearer a001ecb9-05d6-4878-a5af-1ffec64ade40' | jq
[
  {
    "id": "9c414291-8448-458a-8d78-35af2282a394",
    "hash": [
      {
        "hash": "53c84fe23c54bb9dccf549ea495bc7c6",
        "algorithm": "md5"
      },
      {
        "hash": "9a292769693fc1f8b83d5e2f5d087daad31d7ef2ffb268024427af2c5ce6f7f8",
        "algorithm": "sha256"
      }
    ],
    "name": "tonytranscode156",
    "size": 991674,
    "status": "ready",
    "userId": "2f4daa89-4e58-4ee7-bd32-186fb05500ce",
    "createdAt": 1651595161052,
    "updatedAt": 1651595178517,
    "videoSpec": {
      "format": "mp4",
      "tracks": [
        {
          "fps": 30.033333333333335,
          "type": "video",
          "codec": "h264",
          "width": 480,
          "height": 270,
          "bitrate": 96352,
          "duration": 30.5066
        },
        {
          "type": "audio",
          "codec": "aac",
          "bitrate": 160111,
          "channels": 2,
          "duration": 30.5066,
          "sampleRate": 48000
        }
      ],
      "duration": 30.5066
    },
    "playbackId": "9c41n69k4hh3i50h",
    "sourceAssetId": "9a6fc70f-e1b5-4e8b-837c-a6952fe1ad5e",
    "playbackUrl": "https://livepeercdn.monster/recordings/b17134e8-a17b-4d80-9b47-731bb444f015/index.m3u8",
    "downloadUrl": "https://livepeercdn.monster/asset/9c41n69k4hh3i50h/video"
  },
  {
    "id": "9a6fc70f-e1b5-4e8b-837c-a6952fe1ad5e",
    "hash": [
      {
        "hash": "10c918b1d01aea85864ee65d9e0c2305",
        "algorithm": "md5"
      },
      {
        "hash": "8ab080c1406dff77f8897955cf977e9ad779e40ab3a07bc2f8694fbd2fc2be21",
        "algorithm": "sha256"
      }
    ],
    "name": "tonytest1",
    "size": 9840497,
    "status": "ready",
    "userId": "2f4daa89-4e58-4ee7-bd32-186fb05500ce",
    "createdAt": 1651595072766,
    "updatedAt": 1651595089059,
    "videoSpec": {
      "format": "mp4",
      "tracks": [
        {
          "fps": 30,
          "type": "video",
          "codec": "h264",
          "width": 1280,
          "height": 720,
          "bitrate": 2453499,
          "duration": 30.033333,
          "pixelFormat": "yuv420p"
        },
        {
          "type": "audio",
          "codec": "aac",
          "bitrate": 160000,
          "channels": 2,
          "duration": 30.526667,
          "sampleRate": 48000
        }
      ],
      "duration": 30.526667
    },
    "playbackId": "9a6fw3cmw5mzzywe",
    "playbackUrl": "https://livepeercdn.monster/recordings/2ffd5555-d2a3-4ac7-aedd-41072aa92d27/index.m3u8",
    "downloadUrl": "https://livepeercdn.monster/asset/9a6fw3cmw5mzzywe/video"
  }
]
```

TODO:
Update `ReportProgress` to include `prepare` task
Improve efficiency by refactoring uploading/downloading the files
Check output file to make sure full file is there / correct


task/prepare.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/livepeer/go-livepeer/drivers"
+	"github.com/livepeer/stream-tester/model"
+	"github.com/livepeer/stream-tester/segmenter"
+)
+
+func RecordStream(ctx context.Context, lapi *livepeerAPI.Client, playbackID string, osSess drivers.OSSession) (string, error) {
+	glog.Infof("made it into record function")
+	fullPath := videoFileName(playbackID)
+	glog.Infof("full path %s", fullPath)
+	fir, err := osSess.ReadData(ctx, fullPath)
```
hjpotter92, line 20: what does `fir` stand for here?

tqian1: https://github.com/livepeer/task-runner/blob/main/task/transcode.go#L77

`FileInfoReader`

Go's insistence on extra short variable names is bizarre to me 😂

victorges: I think that's more coding style than go's general pattern. I'm down for using more descriptive names everywhere 👀

task/import.go
```diff
@@ -59,6 +61,12 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
 		// TODO: Delete the uploaded file
 		return nil, err
 	}
+	// RecordStream (on our file) after async goroutines finish
```
victorges, line 64: Is this accurate?
```suggestion
	// RecordStream (from our imported file) after async goroutines finish
```

task/prepare.go
victorges, line 20: Would be cool to make this function receive an `io.Reader` instead and just run it in parallel with the regular `Probe` and `Save` tasks in the `Import` function. Not a must though.

tqian1: Eli mentioned it is idiomatic in Go to use shorter/short as possible variable names, this article makes the distinction to do so proportionately to scope which is insightful 
https://murfitt.net/blog/short-variable-names-in-go

task/prepare.go
```diff
@@ -0,0 +1,80 @@
...
+		return "", err
+	}
+	defer sourceFile.Close()
+	var sourceFileSize int64
+	if fir.Size != nil {
+		sourceFileSize = *fir.Size
+	}
+
+	streamName := fmt.Sprintf("vod_recording_%s", time.Now().Format("2006-01-02T15:04:05Z07:00"))
+	stream, err := lapi.CreateStreamEx(streamName, true, nil)
```
victorges, line 37: Might want to use `CreateStreamEx2` to set custom profiles instead. I say we process the info from the `asset` here, especially resolution, to create a specific list of profiles that we want the asset in. Could be as simple as sending the regular 240p, 480p, 720p, 1080p (the transcoders will handle keeping the aspect ration in the response), but removing any profiles with a resolution and/or bitrate higher than the current asset. WDYT?

task/prepare.go
```diff
@@ -0,0 +1,80 @@
...
+
+	gctx, gcancel := context.WithCancel(ctx)
+	defer gcancel()
+	segmentsIn := make(chan *model.HlsSegment)
+	if err = segmenter.StartSegmentingR(gctx, sourceFile, true, 0, 0, segLen, false, segmentsIn); err != nil {
+		return "", err
+	}
+	ws := fileWriter(sourceFileSize)
+	defer ws.Close()
+	var transcoded [][]byte
```
victorges, line 51: I think these are unused.

task/prepare.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		glog.V(model.VERBOSE).Infof("Transcode %d took %s\n", len(transcoded), time.Since(started))
+	}
+	if err == io.EOF {
+		err = nil
+	}
+	if err != nil {
+		return "", err
+	}
+
```
victorges, line 78: I think it would be nice to check if the recording is correct here. We can start by just checking if it has the right duration. If it is not, we should make sure to fail the task instead of creating the asset with broken playback.

task/prepare.go
```diff
@@ -0,0 +1,80 @@
...
+	return stream.ID, nil
```
victorges, line 79: So here you're not actually looking for the stream ID. What you actually want is the **child** stream ID. So check the `record-tester` imlpementation on how it does to grab a recording URL. From what I can remember, it calls an API to list the child sessions of a stream and then picks the only session that is available. If there is more than 1 (or no) sessions available, it is also certainly an error. So you have to check on that here as well and retur the session (child stream) ID instead.

task/prepare.go
victorges: Right, so I think I also do use shorter names when it is a method receiver or very common arguments like `*http.Request`, `http.ResponseWriter`, reader, writer etc. I don't think it is the case here since I don't have `FileInfoReaders` all around I guess. Although it **is** a reader and is also only used close to where it is declared anyway, so not sure... I guess it's kinda subjective in the end. This quote as well:

> More unusual things and global variables need more descriptive names.

IMO if you need an IDE to know what is a thing or keep having to go back to its definition (enters that "scope" insight), you might as well use a more descriptive name. In this case, IMO it's not obvious that `ReadData` returns something called `FileInfoReader`, so a descriptive name would help. If it were called `OpenFileReader`, `ReadFile` or sth like that, then it could make more sense.

But this is a small nit anyway. I wouldn't pro-actively ask to rename this, but since we're already talking/discussing about it that is my perspective on short var names.

task/prepare.go
tqian1: Right, I was initially calling `lapi.GetSessionsR` with the `stream.ID` here but getting `no sessions available`, also took a look inside the DB and could not find any child streams/sessions being created for this stream. Any insight on how come this might not be happening, is it pertinent only to my local stack or is there a nuance to triggering sessions to be created outside of just streaming segments in and having { record: true } on.

task/prepare.go
tqian1: Good point, will add that

task/prepare.go
tqian1: Yup that makes sense as well, actually for testing I am only importing a 480p file since my computer cannot transcode any higher - so it is wrong to be outputting 1080p, 720p etc

task/prepare.go
tqian1: I would like to do this, though for transcode it would need to run in sequence, so will leave this for later to reduce complexity for now

task/import.go
tqian1: No, I will update the PR to do this however so that the `prepare/record` task works on the `imported` file on our end - similar to the pattern in `transcode`

Forgot to update this comment to the current state of the PR

task/prepare.go
tqian1: Changed it to using the stream.ID because session is not required.

task/prepare.go
tqian1: Can we defer this check to a fast follow-up PR to get HLS playback released sooner? May not be trivial to perform this check unsure how long implementation will take.

task/prepare.go
tqian1: Can we defer this to a follow-up PR as well?

task/prepare.go
victorges: Yeah certainly not a must!

task/prepare.go
victorges: I think it's a pretty weird experience to send in a 480p file and get 720p and 1080p back, especially since they will not have a better quality. As we chatted offline I sent a suggestion of what this logic could look like, I think it wouldn't be that complex.

If we can't get it to work for the deploy now it's ok to start without it tho

task/prepare.go
victorges: Sure

task/prepare.go
victorges: Ah right, I didn't know that we didn't create sessions for these streams made directly to the broadcaster.

task/import.go
```diff
@@ -47,18 +49,39 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
...
+	// Download our imported output file
+	fileInfoReader, err := osSess.ReadData(ctx, fullPath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading imported file from output OS path=%s err=%w", fullPath, err)
+	}
+	importedFile, err := readFile(fileInfoReader)
+	if err != nil {
+		return nil, err
+	}
+	fileInfoReader.Body.Close()
```
victorges, line 77: If the function above errs, this will leak a file descriptor/connection to the object storage. Instead of handling all of those corner cases manually when there are branches in the code with an expensive resource, I prefer to just `defer Close()` right after getting the thing and not worry about it. 

There's no problem with keeping the file open a little longer until this function exits and the damage is lower than if we start having accidental leaks, so I just go with the `defer` pattern most of the time.
```suggestion
	}
	defer fileInfoReader.Body.Close()
	importedFile, err := readFile(fileInfoReader)
	if err != nil {
		return nil, err
	}
```

Bonus: If you create a helper as I suggested above to avoid the need for a code comment, this `defer Close()` would already close the file ASAP, and it would only return the `importedFile` arg which the caller themselves would `defer Close()` instead. It's the most optimal "file handle lifecycle" with the safety of `defer` handling the closing.

task/import.go
```diff
@@ -47,18 +49,39 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
...
+	defer importedFile.Close()
+	// RecordStream on output file for HLS playback
+	playbackRecordingId, err := RecordStream(ctx, tctx.lapi, importedFile)
+	if err != nil {
+		return nil, fmt.Errorf("error preparing imported file err=%w", err)
+	}
+	metadata.AssetSpec.PlaybackRecordingID = playbackRecordingId
```
victorges, line 84: I normally avoid mutating objects for eventual concurrency safety, so we could easily do a copy here instead:
```suggestion
    assetSpec := *metadata.AssetSpec
	assetSpec.PlaybackRecordingID = playbackRecordingId
	// use assetSpec instead
```

in JS this would be that simple sugar-y: 
`const assetSpec = { ...metadata.assetSpec, playbackRecordingID }`

task/prepare.go
```diff
@@ -0,0 +1,121 @@
...
+		Fps:     0,
+		Bitrate: 3000000,
+		Width:   1280,
+		Height:  720,
+		Gop:     "2.0",
+	},
+}
+
+func RecordStream(ctx context.Context, lapi *livepeerAPI.Client, assetSpec *livepeerAPI.AssetSpec, file io.ReadSeekCloser) (string, error) {
+	streamName := fmt.Sprintf("vod_hls_recording_%s", time.Now().Format("2006-01-02T15:04:05Z07:00"))
```
victorges, line 53: Maybe it will be more useful to have the asset ID instead of a timestamp here?
```suggestion
	streamName := fmt.Sprintf("vod_hls_recording_%s", assetId))
```
(just gotta receive the ID arg)

Actually, this should just receive a full `TaskContext` which will have the `ctx`, `lapi` and the `AssetId` already. This is a crucial part of a task, which could even be a task itself, so I think we should make it have a similar interface.

task/prepare.go
```diff
@@ -0,0 +1,121 @@
...
+			break
+		}
+		glog.V(model.VERBOSE).Infof("Transcode %d took %s\n", len(transcoded), time.Since(started))
+	}
+	if err == io.EOF {
+		err = nil
+	}
+	if err != nil {
+		return "", err
+	}
```
victorges, line 96: **nit**
```suggestion
	if err != nil && err != io.EOF {
		return "", err
	}
```

task/transcode.go
```diff
@@ -226,6 +226,23 @@ out:
...
+	// Download our transcoded output file
+	outputFileInfoReader, err := tctx.outputOS.ReadData(ctx, fullPath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading transcoded file from output OS path=%s err=%w", fullPath, err)
+	}
+	transcodedFile, err := readFile(outputFileInfoReader)
+	if err != nil {
+		return nil, err
+	}
+	outputFileInfoReader.Body.Close()
```
victorges, line 238: Same regarding resource leaking.

task/transcode.go
```diff
@@ -226,6 +226,23 @@ out:
...
 	}
+	// Download our transcoded output file
+	outputFileInfoReader, err := tctx.outputOS.ReadData(ctx, fullPath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading transcoded file from output OS path=%s err=%w", fullPath, err)
+	}
+	transcodedFile, err := readFile(outputFileInfoReader)
+	if err != nil {
+		return nil, err
+	}
```
victorges, line 237: Now with https://github.com/livepeer/task-runner/pull/16/commits/d46c3ec70746c63d7525a838567717816caae23f we don't need to re-download the file here. Can just use `.Reader()` again right?

task/transcode.go
```diff
@@ -226,6 +226,23 @@ out:
...
+	outputFileInfoReader.Body.Close()
+	defer transcodedFile.Close()
+	// RecordStream on output file for HLS playback
+	playbackRecordingId, err := RecordStream(ctx, tctx.lapi, metadata.AssetSpec, transcodedFile)
+	if err != nil {
+		return nil, fmt.Errorf("error preparing imported file err=%w", err)
+	}
+	metadata.AssetSpec.PlaybackRecordingID = playbackRecordingId
```
victorges, line 245: Same regarding not mutating objects (still optional though)

victorges (approved):
LGTM!

task/import.go
```diff
@@ -47,18 +49,39 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
...
+		return nil, err
+	}
+	fileInfoReader.Body.Close()
+	defer importedFile.Close()
+	// RecordStream on output file for HLS playback
+	playbackRecordingId, err := RecordStream(ctx, tctx.lapi, metadata.AssetSpec, importedFile)
+	if err != nil {
+		glog.Errorf("error preparing imported file err=%w", err)
+	}
+	metadata.AssetSpec.PlaybackRecordingID = playbackRecordingId
```
victorges, line 84: Now that the error is only logged, we need to:
 - fix formatting of error and include contextual information in the log
 - ignore the result if the error is not nil
```suggestion
		glog.Errorf("Error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
	} else {
		metadata.AssetSpec.PlaybackRecordingID = playbackRecordingId
	}
```

victorges (approved):
LLGGTTMM!!

PR #17, opened 2022-04-13
task/probe: Disallow MOV files from being imported

victorges:


victorges:
Closing this as we already disallow MOV from a joy4 incompatibility. This will also be fixed when we move to catalyst soon.

Issue #19, opened 2022-04-29
Import tasks getting stuck mid execution


victorges:
We have some weird cases of tasks getting stuck mid execution. These tasks just keep getting
re-executed, since the RabbitMQ will eventually timeout (after 30m) waiting on an ACK and just
terminates the connection with the client (nacking all messages it had in-flight). The task is then
re-executed as if nothing happened [1].

> [1] This is also another bug that we should address. Should just fail if the task had already tried
running before and just disappeared, which we can already tell from the metadata in the API. This 
is not the root cause though so we still need to investigate and fix the stuck tasks.

No logs that indicate what is wrong, but I have a light suspicion on either:
 - the "progress reporting" logic
 - the "piping" logic in the import task which sends a stream both to `ffprobe` and to the storage
 - the S3 upload client

On the first tasks I found this error, they were actually importing large stream recordings which take
12+ minutes to download on a good connection, due to the on-demand MP4 generation bottleneck.
It was already weird since we have [a hard timeout of 10 minutes](https://github.com/livepeer/task-runner/blob/a8bbcae75fdeb1c4a1bf5bbc4cb4088a06c73610/task/runner.go#L21) so the task runner
should have just failed the task, instead of gone silent.

Right now I just found an even weirder case though. It was from a regular "import" task, which is not
importing a recording but actually just another asset as a test that the user was making. This is the
task:
```json
{
    "id": "51ea2a1e-618e-452d-a024-7c5a0ace266f",
    "type": "import",
    "params": {
        "import": {
            "url": "https://livepeercdn.com/asset/REDACTED/video"
        }
    },
    "status": {
        "phase": "running",
        "progress": 0.649,
        "updatedAt": 1651269956139
    },
    "userId": "REDACTED",
    "createdAt": 1650886712179,
    "outputAssetId": "4582de3b-ead3-4ffe-8b6d-b130f61290a1"
}
```

The asset has around 5GB and takes less than a minute to download from a good connection, so there's
no clear reason of why the task-runner is getting stuck.

PR #20, opened 2022-04-29
task/runner: Avoid re-running poison tasks

victorges:
We have some tasks that are being re-executed
over and over again since they get stuck in the
task-runner logic. We should fix the root cause
of those, but to avoid the problem from getting
worse we should also avoid re-running these tasks
over and over again.

This fixes that by not even starting tasks that we
find out had already been started (phase=running).

This is related to #19

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -144,9 +144,12 @@ func (r *runner) handleTask(ctx context.Context, taskInfo data.TaskInfo) (output
 
 	handler, ok := r.TaskHandlers[strings.ToLower(taskType)]
 	if !ok {
-		return nil, UnretriableError{fmt.Errorf("unknown task type=%q id=%s", taskType, taskID)}
+		return nil, UnretriableError{fmt.Errorf("unknown task type=%q", taskType)}
 	}
 
+	if taskCtx.Status.Phase == "running" {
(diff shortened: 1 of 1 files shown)


tqian1:
Interesting, from a quick glance - is there concept of moving to a DQL (dead letter queue / exchange) after certain amount of retries in RMQ? This will still cycle the task through the queue over and over if i understand correctly

victorges:
> Interesting, from a quick glance - is there concept of moving to a DQL (dead letter queue / exchange) after certain amount of retries in RMQ?

Oooh, good call! We should configure that as well in our task queue. In RabbitMQ it's called a "dead letter exchange", from which we can configure a queue to hold all the poison messages. We also use quorum queues, so we can use this `delivery-limit` policy on them to automatically dead-letter the message (or drop, if we don't want DLX) once it hits it:
https://www.rabbitmq.com/quorum-queues.html#poison-message-handling

> This will still cycle the task through the queue over and over if i understand correctly

It will not! Returning an error on that point of the code will still ACK the message on RabbitMQ since it is the calling function that handles that. It will send the task result message with the error and ack the current message. The API will handle the task result msg and fail the task explicitly in the API (which is a little better than dead-lettering as well, so that we have the proper metadata in the API about the task).

PR #23, opened 2022-05-09
task/probe: Support files with more than 2 tracks

victorges:
The only restriction we really have is that it should
have only 1 video track.

Just tried uploading a file with stereo audio and it
didn't work because of that check.


gioelecerati (approved):
LGTM

PR #29, opened 2022-05-12
prepare: effectiveProfile to avoid dimensions shorter than 145 pixels

### LINKED ISSUE #24: Fix "invalid argument" error for some assets transcode
This also fails the `prepare` step, meaning these assets do not get a `playbackUrl`. This also meant we couldn't make the HLS playback generation errors fatal, meaning assets currently have an inconsistent experience where sometimes they have HLS playback URLs and sometimes they don't.

This file seems to reproduce it consistently: [jellyfish.mp4](https://ipfs.livepeer.com/ipfs/Qmc7S7r8SmYmzsNZx2ao5EZESeLc1HvzJqdr5X7WLso4nL)

@cyberj0g [said](https://discord.com/channels/423160867534929930/971483957545816084/972047866707185686) that he had no issues streaming it directly to a broadcaster, which is a good sign since it means it's hopefully not some deep go-livepeer bug that we are hitting. So the fix should probably be in the `task-runner` logic on how we demux the file or the metadata that we send to the broadcaster on the HTTP push.

## Product Questions

1) How much effort do we estimate to investigate and fix?

I'd say it will hopefully take 1-2 days of investigation.

2) What's the risk if we don't fix it, how does the customer experience suffer, etc.

This is probably a [P3 bug](https://www.notion.so/livepeer/Bug-Prioritization-Methodology-DRAFT-e1e43628698740b8966c09b23b5d73e6), being low impact because it only affects some specific assets and thus some specific users (although we don't really know how common it is) and it is high severity because it breaks HLS playback which would break any NFT that we mint out of it with an HLS player.

If the impact is high, meaning a lot of assets in the wild hit this bug, this could as well be a P1.

gioelecerati:
Fixes #24

### DIFF
--- a/task/prepare.go
+++ b/task/prepare.go
@@ -15,6 +15,7 @@ import (
 const (
 	minVideoBitrate         = 100_000
 	absoluteMinVideoBitrate = 5_000
+	minVideoDimensionPixels = 145
 )
 
 var allProfiles = []api.Profile{
@@ -125,8 +126,10 @@ func getPlaybackProfiles(assetVideoSpec *api.AssetVideoSpec) ([]api.Profile, err
 		return nil, fmt.Errorf("no video track found in asset spec")
 	}
 	filtered := make([]api.Profile, 0, len(allProfiles))
-	for _, profile := range allProfiles {
-		if profile.Height <= video.Height && profile.Bitrate < int(video.Bitrate) {
+	for _, baseProfile := range allProfiles {
+		profile := effectiveProfile(baseProfile, video)
+		if profile.Height <= video.Height && profile.Bitrate < int(video.Bitrate) &&
+			profile.Height > minVideoDimensionPixels && profile.Width > minVideoDimensionPixels {
 			filtered = append(filtered, profile)
 		}
 	}
@@ -134,6 +137,15 @@ func getPlaybackProfiles(assetVideoSpec *api.AssetVideoSpec) ([]api.Profile, err
 		return []api.Profile{lowBitrateProfile(video)}, nil
 	}
 	return filtered, nil
+}
+
+func effectiveProfile(profile api.Profile, video *api.AssetTrack) api.Profile {
+	if video.Width >= video.Height {
+		profile.Height = profile.Width * video.Height / video.Width
+	} else {
+		profile.Width = profile.Height * video.Width / video.Height
+	}
+	return profile
 }
 
 func lowBitrateProfile(video *api.AssetTrack) api.Profile {
--- a/task/prepare.go
+++ b/task/prepare.go
@@ -15,6 +15,7 @@ import (
 const (
 	minVideoBitrate         = 100_000
 	absoluteMinVideoBitrate = 5_000
+	// NVIDIA cards with Turing architecture enforce a minimum width/height of 145 pixels on transcoded videos
 	minVideoDimensionPixels = 145
 )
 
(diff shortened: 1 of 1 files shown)


victorges:
> Were you able to do some testing ie. with the Jellyfish video? Would be helpful to add to the PR description.

AFAIK the `transcode-cli` worked with a 258p profile and failed with a 257p profile, which is exactly the theorical threshold to get the effective width to be 145px or 144px.

> This is not the fix for the E2E testing video correct? Sounds like we are still unsure on root cause for that.

Correct! The fix will likely be on the orchestrators instead, since we were just able to reproduce that the issue only started happening on the version 0.5.30. On go-livepeer@0.5.29 it works, and with this fix we should theoretically have no assets failing transcode!

task/prepare.go
```diff
@@ -114,8 +115,10 @@ func getPlaybackProfiles(assetVideoSpec *livepeerAPI.AssetVideoSpec) ([]livepeer
 		return nil, fmt.Errorf("no video track found in asset spec")
 	}
 	filtered := make([]livepeerAPI.Profile, 0, len(allProfiles))
-	for _, profile := range allProfiles {
-		if profile.Height <= video.Height && profile.Bitrate < int(video.Bitrate) {
+	for _, baseProfile := range allProfiles {
+		profile := effectiveProfile(baseProfile, video)
+		if profile.Height <= video.Height && profile.Bitrate < int(video.Bitrate) &&
+			profile.Height > minVideoDimensionPixels && profile.Width > minVideoDimensionPixels {
```
victorges, line 121: These can actually be `>=`!
```
profile.Height >= minVideoDimensionPixels && profile.Width >= minVideoDimensionPixels
```

victorges, line 121: WDYT of actually breaking up these checks just so it's more readable? Can either create temporary variables giving names to the conditions or have multiple `if` checks with `continue` statements here. e.g. WDYT?
```suggestion
        lowerQualityThanSrc := profile.Height <= video.Height && profile.Bitrate < int(video.Bitrate)
        compliesToMinSize := profile.Height >= minVideoDimensionPixels && profile.Width >= minVideoDimensionPixels
		if lowerQualityThanSrc && compliesToMinSize {
```

task/prepare.go
```diff
@@ -15,6 +15,7 @@ import (
 const (
 	minVideoBitrate         = 100_000
 	absoluteMinVideoBitrate = 5_000
+	minVideoDimensionPixels = 145
```
victorges, line 18: Maybe we can add a very small doc here explaining where this comes from.
```suggestion
    // NVIDIA cards with Turing architecture enforce a minimum width/height of 145 pixels on transcoded videos
	minVideoDimensionPixels = 145
```

victorges (approved):
LGTM with some nits and a very small change! (`>`➡️`>=`)

victorges:
@tqian1 Just rebased this on top of `main` and set the `dev` branch to point to it!

Meaning that, if we've already found all the issues, there should be no `invalid argument` happening in staging anymore!

victorges:
It worksssss

https://lvpr.tv/?monster&recording=d7b8451f-4ddc-4e07-931a-8f6860ea5569

figintern (approved):

gioelecerati:
> Actually seems like the most reliable fix for this will actually be in `lpms`, and they already have a fix in-flight! Should we close this one?
> 
> [livepeer/lpms#326](https://github.com/livepeer/lpms/pull/326)

Yes, if it's fixed there there's no need for this hardcoded fix

victorges:
The fix in `lpms` is not yet merged, and I believe we need this to be able to make the prepare task fatal (#40). As such, I'm reopening, merging, and will be deploying this tomorrow so we don't get so many valid videos that would fail in our processing.

This should be reverted as soon as the `lpms` fix is merged and deployed on our Os.

PR #31, opened 2022-05-13
Added Content-Resolution to prepare & transcode

### LINKED ISSUE #25: Fix source resolution in HLS playback
Since we are not setting the `Content-Resolution` header in the HTTP push request, the generated HLS playlist for the assets have a `0x0` in the source track, which also makes the [playback experience](https://lvpr.tv/?monster&recording=f8d28438-597e-45ca-9019-94035d484686) really weird:

Screen Shot 2022-05-10 at 17 20 50

The `0p` resolution is actually the highest quality one (4k). If a user picked a resolution manually and wanted the highest quality they'd likely end up watching the 720p resolution instead and miss all the 4k glory.

refs:
 - [where the header is read](https://github.com/livepeer/go-livepeer/blob/911920ee5e18b544864f08a9ecfedb6d4d479cd0/server/mediaserver.go#L774)
 - [where the header is not set](https://github.com/livepeer/stream-tester/blob/4776d7656c424c1d982869baa2d1ae0c38e25980/apis/livepeer/livepeer.go#L864)

## Product Questions

1) How much effort do we estimate to investigate and fix?

Half a day work.

This should be a really simple fix since we already have the asset resolution, it's just a matter of passing it around.

2) What's the risk if we don't fix it, how does the customer experience suffer, etc.

This is probably a [P2 bug](https://www.notion.so/livepeer/Bug-Prioritization-Methodology-DRAFT-e1e43628698740b8966c09b23b5d73e6), being high impact because it happens with every asset but low severity because it does not disrupt the playback experience completely.

gioelecerati:
Fixes #25 
Depends on https://github.com/livepeer/go-api-client/pull/6  

**What does it do?**
Retrieve the source asset resolution and provide it to PushSegment, to be used as a Content-Resolution header and display the source resolution correctly.


task/transcode.go
```diff
@@ -161,6 +161,13 @@ func TaskTranscode(tctx *TaskContext) (*data.TaskOutput, error) {
 	outFiles = append(outFiles, mp4muxer)
 	outBuffers = append(outBuffers, ws)
 	var transcoded [][]byte
+	contentResolution := ""
+	for _, track := range tctx.InputAsset.AssetSpec.VideoSpec.Tracks {
+		if track.Type == "video" {
+			contentResolution = fmt.Sprintf("%dx%d", track.Width, track.Height)
+			break
+		}
+	}
```
victorges, line 170: **nit:** We should probably create a litle helper/utility to grab the video track from an asset spec (or maybe receiving a `trackType string` and returning the first track with that). We have this exact same loop repeated in a couple of places now.

task/transcode.go
```diff
@@ -161,6 +161,13 @@ func TaskTranscode(tctx *TaskContext) (*data.TaskOutput, error) {
 	outFiles = append(outFiles, mp4muxer)
 	outBuffers = append(outBuffers, ws)
 	var transcoded [][]byte
+	contentResolution := ""
```
victorges, line 164: Should check either here or in the client that the `contentResolution` value is not empty (or if we change it to width/height, that it is `> 0`). We shouldn't set the header with an empty value.

Even though we theoretically don't have asses without video tracks / resolution info right now, this could change in the future and it's good to have the code reliable to that.

victorges (approved):

figintern (approved):
LGTM!

task/transcode.go
victorges: Do you know if the broadcaster would support a value like that? If so, that could be cool :)

PR #35, opened 2022-05-25
clients/ipfs: Use CIDv1 for all IPFS uploads

victorges:
It is the recommended approach for CIDs as it's
future-proof (hash alg is embedded in the CID)
and also because it's case insensitive and will
work better on all scenarios (like URLs).

It is weirdly not the default on Piñata yet, but
while it is not we can easily toggle it on our
export requests.

It does not break compatibility on anything that I can
think of.

Reference: https://docs.ipfs.io/how-to/best-practices-for-ipfs-builders/

### DIFF
--- a/clients/ipfs.go
+++ b/clients/ipfs.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


victorges:
FTR just minted this to make sure it works fine E2E with the CIDv1
https://testnets.opensea.io/assets/mumbai/0xA4E1d8FE768d471B048F9d73ff90ED8fcCC03643/84

gioelecerati (approved):
LGTM

Issue #33, opened 2022-05-16
NFT metadata templating


victorges:
To allow creation of Lens publication as well as regular video NFTs. For that we need to implement some kind of NFT metadata templating in the `task-runner`.
  - This is to allow minting both regular NFTs for marketplaces but also the Lens publications which have additional fields (including some “dynamic” fields, using the CID of the file exported to IPFS).
  - We can implement this in a couple of ways, here are some ideas:
      * A full templating language on the requests, setting fields to something like `{{ .output.ipfs.videoFileUrl }}`, or
      * Just a set of pre-defined templates in the code (e.g. right now `erc721` or `lens_publication`). We just specify on the request which template to use. This will give us less flexibility but should be much simpler to implement, so may be preferred.

https://www.notion.so/livepeer/Lens-Integration-e756c036e6c741de97d8e9ef1c351e84#6394c0df8bdc464f9285aecb458161f3

Defintions of done:
    1. Can export to IPFS a metadata corresponding to a Lens publication, following all the spec described here
    2. Can also export to IPFS a metadata corresponding to a regular Video NFT without any of the Lens-specific fields

PR #36, opened 2022-05-28
task/export: Implement NFT metadata templating

victorges:
Right now the templating is used to create an NFT metadata either with the player as the `animation_url`
or with the raw file. 

This will be especcially necessary for supporting the Lens publications NFTs as well.
For context: [[design doc]](https://www.notion.so/livepeer/Lens-Integration-Engineering-Spec-e756c036e6c741de97d8e9ef1c351e84#90373644024348108d2852ecaa21aa6f)

This implements #33


gioelecerati (approved):
LGTM

PR #38, opened 2022-06-06
task/progress: Add Progress Tracking to Transcode and Prepare

### LINKED ISSUE #26: Update task progress on transcode and prepare
Running a transcode right now is kinda of a black box since you get no update on how far is it in the process. As the prepare sub-task was implemented based on the same code, it also does not make any progress updates along the way and the user may stay a minute or 2, or even longer, without any feedback from the system. 

This is problematic especially because these tasks take so long and so by getting no feedback the user might as well assume that it didn't work and give up. When I got that behavior, I actually went on the logs to make sure that something was actually happening and it wasn't just stuck. Regular users cannot do that.

We also already have some abstractions in the code to do the progress reporting already, which are used by the import and export tasks: https://github.com/livepeer/task-runner/blob/071d23ce473372af07485a2b6b44d321dd113f02/task/progress.go#L19

We could very easily use that from the transcode task.

For prepare, what we would need is some logic for updating the progress partially (maybe only until 50%) on the main import/transcode tasks, and then update the other 50% in the prepare sub-task. We can implement it with some simple `start/end` parameters to the `ReportProgress` helper, or some more complex abstraction if it's worth it.

figintern:
https://github.com/livepeer/task-runner/issues/26 https://github.com/livepeer/task-runner/issues/40

Refactor `ReportProgress` logic to specify progress sub-window between 0 and 100 percent. Percentage progress of getCount/size is applied only to the sub-window of overall progress.

Track progress for `transcode` and `process` tasks using size of segments pushed relative to overall size of the file.

`transcode`/`process` and `import`/`process` tasks are 50-50 split between the progress bar, this is easily adjustable.

---

Makes Prepare sub-task failures fatal

---

Also fixes bug in `probe.go` which causes error when `video pixel format` could not be probed from file and is empty

### DIFF
--- a/task/export.go
+++ b/task/export.go
@@ -47,7 +47,7 @@ func TaskExport(tctx *TaskContext) (*data.TaskOutput, error) {
 	tctx = tctx.WithContext(ctx)
 
 	content := NewReadCounter(file.Body)
-	go ReportProgress(ctx, tctx.lapi, tctx.Task.ID, size, content.Count)
+	go ReportProgress(ctx, tctx.lapi, tctx.Task.ID, size, content.Count, 0, 100)
 	output, err := uploadFile(tctx, asset, content)
 	if err != nil {
 		return nil, err
--- a/task/export.go
+++ b/task/export.go
@@ -47,7 +47,7 @@ func TaskExport(tctx *TaskContext) (*data.TaskOutput, error) {
 	tctx = tctx.WithContext(ctx)
 
 	content := NewReadCounter(file.Body)
-	go ReportProgress(ctx, tctx.lapi, tctx.Task.ID, size, content.Count, 0, 100)
+	go ReportProgress(ctx, tctx.lapi, tctx.Task.ID, size, content.Count, 0, 1)
 	output, err := uploadFile(tctx, asset, content)
 	if err != nil {
 		return nil, err
--- a/task/import.go
+++ b/task/import.go
@@ -137,9 +137,7 @@ func prepareImportedAsset(tctx *TaskContext, metadata *FileMetadata, fullPath st
 
 	playbackRecordingID, err := Prepare(tctx, metadata.AssetSpec, importedFile, 0.5)
 	if err != nil {
-		glog.Errorf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
-		// TODO: make these fatal once we're confident about prepare reliability
-		return "", nil
+		glog.Fatalf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
 	}
 	return playbackRecordingID, nil
 }
--- a/task/import.go
+++ b/task/import.go
@@ -9,7 +9,6 @@ import (
 	"mime"
 	"net/http"
 	"path"
-	"strings"
 
 	"github.com/golang/glog"
 	api "github.com/livepeer/go-api-client"
@@ -34,18 +33,15 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
 	secondaryReader, pipe := io.Pipe()
 	mainReader := NewReadCounter(io.TeeReader(contents, pipe))
 
+	progressCtx, cancelProgress := context.WithCancel(ctx)
+	defer cancelProgress()
+	go ReportProgress(progressCtx, tctx.lapi, tctx.Task.ID, size, mainReader.Count, 0, 0.5)
+
 	eg, egCtx := errgroup.WithContext(ctx)
 	var (
 		videoFilePath, metadataFilePath, fullPath string
 		metadata                                  *FileMetadata
 	)
-	// Temporarily skip preparing recorded streams while we figure out a bug in the orchestrators latest version.
-	isInputRecording := strings.HasPrefix(params.URL, "https://livepeercdn.") && strings.Contains(params.URL, "/recordings/")
-	if !isInputRecording {
-		go ReportProgress(egCtx, tctx.lapi, tctx.Task.ID, size, mainReader.Count, 0, 0.5)
-	} else {
-		go ReportProgress(egCtx, tctx.lapi, tctx.Task.ID, size, mainReader.Count, 0, 1)
-	}
 	// Probe the source file to retrieve metadata
 	eg.Go(func() (err error) {
 		metadata, err = Probe(egCtx, tctx.OutputAsset.ID, filename, mainReader)
@@ -70,25 +66,10 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
 		// TODO: Delete the source file
 		return nil, err
 	}
-	playbackRecordingId := ""
-	// TODO: Remove this check and prepare all assets.
-	if !isInputRecording {
-		// Download our imported output file
-		fileInfoReader, err := osSess.ReadData(ctx, fullPath)
-		if err != nil {
-			return nil, fmt.Errorf("error reading imported file from output OS path=%s err=%w", fullPath, err)
-		}
-		defer fileInfoReader.Body.Close()
-		importedFile, err := readFile(fileInfoReader)
-		if err != nil {
-			return nil, err
-		}
-		defer importedFile.Close()
-		// RecordStream on output file for HLS playback
-		playbackRecordingId, err = Prepare(tctx, metadata.AssetSpec, importedFile, 0.5)
-		if err != nil {
-			glog.Errorf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
-		}
+	cancelProgress()
+	playbackRecordingID, err := prepareImportedAsset(tctx, metadata, fullPath)
+	if err != nil {
+		return nil, err
 	}
 	assetSpec := *metadata.AssetSpec
 	assetSpec.PlaybackRecordingID = playbackRecordingID
@@ -154,7 +135,7 @@ func prepareImportedAsset(tctx *TaskContext, metadata *FileMetadata, fullPath st
 	}
 	defer importedFile.Close()
 
-	playbackRecordingID, err := Prepare(tctx, metadata.AssetSpec, importedFile)
+	playbackRecordingID, err := Prepare(tctx, metadata.AssetSpec, importedFile, 0.5)
 	if err != nil {
 		glog.Errorf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
 		// TODO: make these fatal once we're confident about prepare reliability
--- a/task/import.go
+++ b/task/import.go
@@ -41,10 +41,10 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
 	)
 	// Temporarily skip preparing recorded streams while we figure out a bug in the orchestrators latest version.
 	isInputRecording := strings.HasPrefix(params.URL, "https://livepeercdn.") && strings.Contains(params.URL, "/recordings/")
-	if isInputRecording {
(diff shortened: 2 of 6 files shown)


task/progress.go
```diff
@@ -16,7 +16,9 @@ var progressReportBuckets = []float64{0, 0.25, 0.5, 0.75, 1}
 const minProgressReportInterval = 10 * time.Second
 const progressCheckInterval = 1 * time.Second
 
-func ReportProgress(ctx context.Context, lapi *livepeerAPI.Client, taskID string, size uint64, getCount func() uint64) {
+func ReportProgress(ctx context.Context, lapi *livepeerAPI.Client, taskID string, size uint64, getCount func() uint64, startPercentage int, endPercentage int) {
```
hjpotter92, line 19: small edit: the params can be:

```go
..., startPercentage, endPercentage int)
```
go will assign int type for the `startPercentage`.

task/progress.go
```diff
@@ -16,7 +16,9 @@ var progressReportBuckets = []float64{0, 0.25, 0.5, 0.75, 1}
...
+	startPercentage = int(math.Max(float64(startPercentage), 0)) // >= 0, <= 100, smaller than endPercentage
+	endPercentage = int(math.Min(float64(endPercentage), 100))   // >= 0, <= 100, larger than startPercentage
```
hjpotter92, line 21: should there be a check for `start <= end`?

task/prepare.go
```diff
@@ -52,7 +52,7 @@ var allProfiles = []livepeerAPI.Profile{
 	},
 }
 
-func Prepare(tctx *TaskContext, assetSpec *livepeerAPI.AssetSpec, file io.ReadSeekCloser) (string, error) {
+func Prepare(tctx *TaskContext, assetSpec *livepeerAPI.AssetSpec, file io.ReadSeekCloser, size int64, reportProgressStartPercentage int) (string, error) {
```
victorges, line 55: **nit:** `assetSpec` already has the `size` afaik, so we could skip the extra `size` var

victorges: Also, an idea to avoid the long var name
```suggestion
func Prepare(tctx *TaskContext, assetSpec *livepeerAPI.AssetSpec, file io.ReadSeekCloser, size int64, progressStart int) (string, error) {
```

task/progress.go
victorges: WDYT of making these floating points instead from 0 to 1, since that's the format of the output of this function? Integers only exist here for the count and total size, but then are immediately converted into floats for the percentage calculation.

It will also make some of the code a little simpler, avoiding a lot of `int()` and `float64()` conversions here.

task/progress.go
victorges, line 21: **nit:** I think the comments are unnecessary since the code is relatively simple/self-explanatory (they're also innacurate since we're only enforcing 1 of the 3 things it mentions)

task/progress.go
```diff
@@ -50,8 +52,10 @@ func ReportProgress(ctx context.Context, lapi *livepeerAPI.Client, taskID string
 	}
 }
 
-func calcProgress(count, size uint64) (val float64) {
+func calcProgress(count, size uint64, startPercentage int, endPercentage int) (val float64) {
 	val = float64(count) / float64(size)
+	val = val * (float64(endPercentage) - float64(startPercentage))
+	val = (val + float64(startPercentage)) / 100
```
victorges, line 58: Yeah I think the math here will be simpler with `0-1` floats as well. Sth like this?

```suggestion
	val = startPercentage + val * (endPercentage - startPercentage)
```

task/transcode.go
```diff
@@ -115,6 +116,24 @@ func fileWriter(size int64) WriteSeekCloser {
 	}
 }
 
+type SegmentCounter struct {
```
victorges, line 119: **nit:** Perhaps name this `SegmentLengthCounter` (or size) otherwise it seems like it's just counting the number of segments seen, not their size.

task/transcode.go
```diff
@@ -115,6 +116,24 @@ func fileWriter(size int64) WriteSeekCloser {
...
+	size  uint64
+	count uint64
```
victorges, line 121: **nit:** Would also avoid `count` here not to seem like it's counting the number of segments. A suggestion:
```suggestion
	totalSize  uint64
	readSize uint64
```
(or can even drop the `Size` suffix if the struct name is clear enough)

task/transcode.go
```diff
@@ -115,6 +116,24 @@ func fileWriter(size int64) WriteSeekCloser {
...
+
+func NewSegmentCounter(size int64) *SegmentCounter {
+	return &SegmentCounter{size: uint64(size)}
+}
+
+func (c SegmentCounter) Count() uint64 {
+	return atomic.LoadUint64(&c.count)
+}
+
+func (c SegmentCounter) Read(data []byte) int {
```
victorges, line 132: So this function on the `ReadCounter` was only called this because it is made for wrapping an `io.Reader` into another `Reader`, and so it must follow that interface. Here, the `Read` name seems inaccurate and we don't need to return anything either, so we don't really need to make a similar interface.

So WDYT of actually making this an `Accumulator` instead which only receives an `uint64` and adds to its internal state var? We also:
 - Don't need any `size` variable here. It is not used in its internal logic and only constructed with it to be read once on the next line, so we don't really need it.
 - Can use the accumulator from `ReadCounter` and avoid the `atomic` logic duplication! You can even make it an anonymous field (go's most similar way of "inheriting") of `ReadConter`, so the `Count` function will already be exposed and you only need to call `Add` (or `Accumulate`, or sth like that) from its `Read` method.

task/prepare.go
```diff
@@ -87,6 +87,9 @@ func Prepare(tctx *TaskContext, assetSpec *livepeerAPI.AssetSpec, file io.ReadSe
 		}
 	}
 
+	counter := NewSegmentCounter(size)
+	go ReportProgress(ctx, lapi, tctx.Task.ID, counter.size, counter.Count, reportProgressStartPercentage, 100)
```
victorges, line 91: Same regarding `counter.size`

task/import.go
```diff
@@ -82,7 +86,7 @@ func TaskImport(tctx *TaskContext) (*data.TaskOutput, error) {
 		}
 		defer importedFile.Close()
 		// RecordStream on output file for HLS playback
-		playbackRecordingId, err = Prepare(tctx, metadata.AssetSpec, importedFile)
+		playbackRecordingId, err = Prepare(tctx, metadata.AssetSpec, importedFile, *fileInfoReader.Size, 50)
```
victorges, line 89: `fileInfoReader` not always has a `Size`, which is why it is a pointer instead of a direct value.

One way would be to check before using, but I think it'll be much simpler to just grab the size from `metadata.AssetSpec` (so then you don't even need the separate arg, just grab it from inside `Prepare`)

victorges (approved):
LGTM with a couple of nits. Still hunting for the bug you mentioned though 🧐

task/export.go
```diff
@@ -47,7 +47,7 @@ func TaskExport(tctx *TaskContext) (*data.TaskOutput, error) {
 	tctx = tctx.WithContext(ctx)
 
 	content := NewReadCounter(file.Body)
-	go ReportProgress(ctx, tctx.lapi, tctx.Task.ID, size, content.Count)
+	go ReportProgress(ctx, tctx.lapi, tctx.Task.ID, size, content.Count, 0, 100)
```
victorges, line 50: Shouldn't this be 0 1?

task/import.go
```diff
@@ -131,11 +135,9 @@ func prepareImportedAsset(tctx *TaskContext, metadata *FileMetadata, fullPath st
 	}
 	defer importedFile.Close()
 
-	playbackRecordingID, err := Prepare(tctx, metadata.AssetSpec, importedFile)
+	playbackRecordingID, err := Prepare(tctx, metadata.AssetSpec, importedFile, 0.5)
 	if err != nil {
-		glog.Errorf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
-		// TODO: make these fatal once we're confident about prepare reliability
-		return "", nil
+		glog.Fatalf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
```
victorges, line 140: These should _not_ be fatal to the whole process. They should only be fatal for the task.

So instead of a fatal log here you should just return the error instead of ignoring it. Otherwise the whole process will crash and every other unrelated task will fail with it as well.

victorges (changes requested):
Some minor change which is required

task/prepare.go
```diff
@@ -64,11 +64,11 @@ func Prepare(tctx *TaskContext, assetSpec *api.AssetSpec, file io.ReadSeekCloser
 	streamName := fmt.Sprintf("vod_hls_recording_%s", assetId)
 	profiles, err := getPlaybackProfiles(assetSpec.VideoSpec)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
 	stream, err := lapi.CreateStream(api.CreateStreamReq{Name: streamName, Record: true, Profiles: profiles})
 	if err != nil {
-		return "", nil
+		return "", err
```
victorges, line 71: Good catch

task/progress.go
```diff
@@ -41,7 +45,8 @@ func ReportProgress(ctx context.Context, lapi *api.Client, taskID string, size u
 				progressBucket(progress) == progressBucket(lastProgress) {
 				continue
 			}
-			if err := lapi.UpdateTaskStatus(taskID, "running", progress); err != nil {
+			scaledProgress := scaleProgress(progress, startFraction, endFraction)
```
victorges, line 48: nit: could just iterate on the same variable like `progress = scaleProgress(progress, start, end)`

task/progress.go
```diff
@@ -57,6 +62,11 @@ func calcProgress(count, size uint64) (val float64) {
 	return
 }
 
+func scaleProgress(progress, startFraction, endFraction float64) (val float64) {
+	val = startFraction + progress*(endFraction-startFraction)
+	return
+}
```
victorges, line 68: I think the named return value is only making it more complex here (in fact caused the last bug?). Wdyt of?

```suggestion
func scaleProgress(progress, startFraction, endFraction float64) float64 {
	return startFraction + progress*(endFraction-startFraction)
}
```

task/transcode.go
```diff
@@ -116,6 +117,22 @@ func fileWriter(size int64) WriteSeekCloser {
 	}
 }
 
+type SegmentSizeAccumulator struct {
```
victorges, line 120: nit: So... Nothing in this interface assumes or required that the thing being accumulated is the segment size. It can be used to accumulate any other value, like for example the read bytes in the case of the read counter.

I believe this be just called Accumulator. It can also be used from the ReadCounter above

task/transcode.go
```diff
@@ -237,10 +259,11 @@ out:
 	if err != nil {
 		return nil, err
 	}
+	cancelProgress()
 	// RecordStream on output file for HLS playback
-	playbackRecordingId, err := Prepare(tctx.WithContext(ctx), metadata.AssetSpec, ws.Reader())
+	playbackRecordingId, err := Prepare(tctx.WithContext(ctx), metadata.AssetSpec, ws.Reader(), 0.5)
 	if err != nil {
-		glog.Errorf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
+		glog.Fatalf("error preparing imported file assetId=%s err=%q", tctx.OutputAsset.ID, err)
```
victorges, line 266: Same. Must not crash the process.

victorges (approved):
LGTM! Some optional nits in remaining comments

task/progress.go
victorges: Sounds good :)

PR #43, opened 2022-06-14
task: Update go-api-client and stream-tester with fixes

victorges:
This updates both stream-tester and go-api-client with the following fixes:
 - https://github.com/livepeer/stream-tester/pull/164
 - https://github.com/livepeer/go-api-client/pull/9
 - https://github.com/livepeer/go-api-client/pull/10

Upon fixing the usage of the libs, I also figured out that cancellation was not being
handled properly in the transcode and prepare tasks. Without https://github.com/livepeer/stream-tester/pull/164 
it actually wouldn't be fixable, since the segmenter didn't handle cancellation, but 
now it does!

~~So this also finally fixes NOT https://github.com/livepeer/task-runner/issues/19 !!~~
Actually figured out that this only fixes cancellation on transcode/prepare. There may
still be something wrong with the import logic :(

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/livepeer/go-livepeer v0.5.31
 	github.com/livepeer/joy4 v0.1.2-0.20220210094601-95e4d28f5f07
 	github.com/livepeer/livepeer-data v0.4.14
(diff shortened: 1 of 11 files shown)


figintern (approved):
LGTM

iameli (approved):
LGTM

Issue #42, opened 2022-06-13
Create log for files with a start time


pglowacky:
Log every time we see a file with a start time

PR #44, opened 2022-06-14
task/probe: Add log with ffprobe output (incl `start_time`)

victorges:
We need this in order to create a better sense of the kind of files that people upload to our API.

We can assume that the start_time bug is not very frequent, but we don't really know until we have this.

After having this, should also create some dashboards and/or alerts to know if too many files with
a big start_time are bing uploaded to the VOD API.

This implements #42


figintern (approved):
LGTM

PR #45, opened 2022-06-14
task/import: Use original HLS recording for assets playback URL

victorges:
For assets created from a livestream recording, we have been skipping the prepare step because of 
an apparent bug in `go-livepeer` that caused them not to be processed properly (`Invalid argument`).

The ideal flow for this though would actually be to avoid re-transcoding these recordings when we are
creating the assets. We have already created an HLS playlist for the livestream when it was recorded
in the first place. So we can just use that for the resulting asset playback as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/golang/glog v1.0.0
-	github.com/livepeer/go-api-client v0.2.2-0.20220614232202-3f52abca6d18
+	github.com/livepeer/go-api-client v0.2.2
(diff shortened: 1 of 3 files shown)


figintern (approved):
My understanding of this PR is that for assets of livestreams with `record = true`, there is a `RecordedSessionID` field that is already populated. Rather than run through `Prepare` redundantly and associate the resulting `playbackRecordingID` from prepare, we populate `PlaybackRecordingID` with the `RecordedSessionID` that already exists and bypass `Prepare`.

victorges:
@tqian1 100% accurate understanding!

PR #46, opened 2022-06-17
task/probe: Fix some remaining bugs in probe

victorges:
- List of supported pixel formats is incomplete
 - We are using a really really old version of ffmpeg (major 3) which doesn't even detect the pix format sometimes. Buster bumps it at least to major 4.
 - Improve some error messages


figintern (approved):
LGTM

PR #47, opened 2022-06-23
Revert "prepare: effectiveProfile to avoid dimensions shorter than 14…

figintern:
…5 pixels (#29)"

This reverts commit 364297c96244f4d250642c3ae4c233e167613438.


victorges (approved):
Have we deployed the fix on lpms yet? If so, LGTM!

PR #52, opened 2022-07-27
probe: Ignore data tracks in assets

### LINKED ISSUE #1157: Bug: Assets uploaded with subtitle tracks fail
Reported by Lenstube that: https://livepeercdn.com/asset/48eeigdrovj7norx/ video is not found.

This was due to the processing of the asset failing due to having a `data` track, which we don't support 

Expected behavior:

As long as assets have a video and audio track, they should succeed. Any additional fields should be ignored rather than failing to process.

victorges:
This is to merely ignore any `data` tracks present in tasks in order to not have them fail processing because of that.

We will likely be just ignoring those, and thus not having the subtitles in the resulting HLS renditions, but that's
still better than just failing the processing altogether and not importing the asset properly to the VOD API.

Also considered still keeping information about the data track in the asset, but preferred not to while we don't
have any real handling of those, just ignore them. Let's not commit to any interfaces about how that's gonna look
in the future.

This fixes livepeer/studio#1157

### DIFF
--- a/task/probe.go
+++ b/task/probe.go
@@ -83,6 +83,8 @@ func toAssetSpec(filename string, probeData *ffprobe.ProbeData, size uint64, has
 		track, err := toAssetTrack(stream)
 		if err != nil {
 			return nil, err
+		} else if track == nil {
+			continue
 		}
 		if track.Type == "video" {
 			if hasVideo {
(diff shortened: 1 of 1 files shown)


gioelecerati (approved):
LGTM

PR #53, opened 2022-08-05
Replace go-livepeer dependency with go-tools

cyberj0g:
Drivers package is now in a separate repo. Task-runner doesn't seem to run any tests on PR, and the tests are failing for me locally for seemingly unrelated reason, could someone please check?


go.mod
```diff
@@ -5,21 +5,14 @@ go 1.16
 require (
 	github.com/golang/glog v1.0.0
 	github.com/livepeer/go-api-client v0.2.8-0.20220729152532-d7db15db3379
-	github.com/livepeer/go-livepeer v0.5.31
```
victorges, line 8: This will make the build a lot better as well! Won't need to install all the ffmpeg dependencies. Could you remove that bit from the `Dockerfile`? If it doesn't work out of the box no worries, i can figure it out later.

https://github.com/livepeer/task-runner/commit/219634e64124efcc8d5bfcf93d5c8cbc9ad36915#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R4-R10

victorges (approved):
Thanks for this! A separate library for this has been long anticipated.

victorges:
> Task-runner doesn't seem to run any tests on PR, and the tests are failing for me locally for seemingly unrelated reason, could someone please check?

That is correct, we do not have unit tests in this project yet. That `import_test` one has been broken for a while, you can leave it as is or just remove it! (probably more confusing to have it there)

Issue #50, opened 2022-07-22
Call DMS for the upload task


victorges:
This is the part that will replace both the `probe.go` and `prepare.go` flows with the [DMS API for VOD & Recording](https://www.notion.so/livepeer/Technical-Design-Studio-DMS-API-for-VOD-Recording-bba2b160efff4ad4963dae2d4e9ff1a2).

PR #54, opened 2022-08-05
tasks: Implement `upload` task calling Catalyst for processing

victorges:
This is to create the proper `Upload` task in `task-runner`, which will replace the `import` task with 
a pipeline solely handled by the new Catalyst VOD funcitonalities.

The biggest change in the business logic of the `task-runner` for this was the creation of a "task step"
abstraction. This is a way of scheduling new messages on RabbitMQ about a given task and is used 
here for scheduling the continuation of a task execution after the Catalyst callback. it will be used later
for a health check on the Catalyst job as well, by scheduling some delayed task steps.

This implements #50


task/upload.go
```diff
@@ -0,0 +1,58 @@
...
+			},
+			{
+				Type:            "ipfs_pinata", // TODO: Add this based on asset.storage
+				PinataAccessKey: tctx.PinataAccessToken,
+			},
+		},
+	}
+	if err := tctx.catalyst.UploadVOD(ctx, uploadReq); err != nil {
+		return nil, fmt.Errorf("failed to call catalyst: %v", err)
+	}
```
victorges, line 41: :happening:

thomshutt:
Looks good so far! Seems to line up with what we've discussed

Issue #49, opened 2022-07-22
Create an API server in task-runner for DMS callback


victorges:
This will be called with the progress of the executing task, which the `task-runner` should proxy to the API, and in the end with the result of the task.

The callback with the result should necessarily generate an event in RabbitMQ. That event can be either:
 - a `task_result` event, just like today, with the result of the task to the API
 - some new event to "continue" the task. this might be the case for IPFS storage for example, in which we might still wanna store the metadata on IPFS after the asset has been stored there.

A possibility is for this API to be as simple as getting the API call and immediately sending an event with the payload without processing anything. This will guarantee the most reliability, assuming we can trust RabbitMQ, and also performance so DMS never has to wait on any processing that we do with the callback.

Notice that this involves some `livepeer-infra` changes as well, since currently there is not even any `Service` in `task-runner` since it only receives requests via RabbitMQ.

PR #56, opened 2022-08-09
api: Create an API server for Catalyst callbacks

victorges:
This is to create an API server that will be called by Catalyst both during the processing of a video
to report progress on the task, as well as when it is finished to report on the final result.

Played around with different possibilities but in the end just kept the handling of the callback inside
`task/runner.go` itself, which is the core of the `task-runner` implementation right now.

Implements #49


api/auth.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+)
+
+func authorized(authSecret string, next http.Handler) http.HandlerFunc {
+	return func(rw http.ResponseWriter, r *http.Request) {
+		authHeader := r.Header.Get("Authorization")
```
thomshutt, line 11: This is something I hadn't actually considered - should we be auth-ing in our callbacks? We're currently just doing a raw POST to whatever URL you pass us

task/runner.go
```diff
@@ -216,6 +217,23 @@ func (r *runner) getAssetAndOS(assetID string) (*api.Asset, drivers.OSSession, e
 	return asset, osSession, nil
 }
 
+func (r *runner) HandleCatalysis(ctx context.Context, taskId, nextStep string, callback *clients.CatalystCallback) error {
```
thomshutt, line 220: Catalysis is a great word

api/auth.go
victorges: Especially since we'll be exposing the `task-runner` API to internet, I think we _should_ have this auth here as well!

If it is too much of a lift to do it on the `catalyst-api` side, I could make an "internal" authorization logic here, like putting the auth secret in the query string. That would be a little less secure overall tho (any service logging the URLs will log the secret too), but would work. So LMK if you think you'll be able to add the secret on the callback or if I should do something like that

Issue #51, opened 2022-07-22
Implement healthcheck of DMS tasks with RabbitMQ delayed msg


victorges:
DMS will get a task to run and it will execute. It is supposed to call back in the provided callback URL
every few seconds to update the progress ("precentage") and also in the end with the result. 

At least in the initial version though, we will have no guarantees about the task being really executed to
the end in case of occasional outages of restarts of the DMS node. To sustain those kinds of failures, we
should make some kind of periodic check of the task status while it has been delegated to DMS.

The idea is:
 - after starting the task in DMS, send a delayed event for +1 minute (example) from now
 - when the event is received, check if either the status of the task has been updated "recently"  (e.g. 30s ago) or if it has already finished. 
 - If task has been updated recently, schedule another event for 1min from now.
 - if the task is finished we are done, nothin to do, just ack the event and exit. 
 - If neither, it means that DMS has gone away so we need to restart the task on their side. Call DMS again and restart the flow.

PR #57, opened 2022-08-09
task/upload: Create healthcheck to track catalyst tasks

victorges:
This creates a way of delaying "task step schedule" events in order to implement a recurring healthcheck
on the catalyst tasks. This means that while Catalyst is executing on some task, the `task-runner` will be 
periodically (every minute) checking if it has received at least a progress callback from Catalyst. If it hasn't
received anything in over a minute, it will consider the Catalyst job to be lost and fail the task on the Studio side.

This implements #51

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -71,9 +71,10 @@ func NewRunner(opts RunnerOptions) Runner {
 		opts.TaskHandlers = defaultTasks
 	}
 	return &runner{
-		RunnerOptions: opts,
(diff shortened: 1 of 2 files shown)


thomshutt (approved):
LGTM, 1 minute should be very conservative but might be worth bumping the timeout to something like 5 mins initially until we have a better picture of how well things work / how frequently we send the various callback.

PR #60, opened 2022-08-16
Fix missing arch in release file names

cyberj0g:


.github/workflows/build.yaml
```diff
@@ -25,7 +25,9 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: 1.17
-
+      - name: Export platform env
+        run: |
+         echo "GOARCH=${{ matrix.arch }}" >> $GITHUB_ENV
```
hjpotter92, line 30: why though?

hjpotter92: nvm, just saw the released archive names. :facepalm:

hjpotter92 (approved):

PR #61, opened 2022-08-16
runner: add basic replace of error text messages

### LINKED ISSUE #1162: Error Handling: Upload Errors
When uploading an asset, a user may encounter an error. Currently the errors provided are unclear to users and don't suggest whether or not the asset should be retried. 

- Need list of all errors we've seen during upload
- Clarify the language passed to users

gioelecerati:
Add a basic replace of error messages with more clear messages sent to Studio.
Fixes [1162](https://github.com/livepeer/studio/issues/1162)

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -221,6 +221,22 @@ func (r *runner) getAssetAndOS(assetID string) (*api.Asset, drivers.OSSession, e
 
 func (r *runner) publishTaskResult(ctx context.Context, task data.TaskInfo, output *data.TaskOutput, err error) error {
 	resultCh := make(chan event.PublishResult, 1)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "unexpected EOF") {
+			err = errors.New("file download failed")
+		} else if strings.Contains(err.Error(), "error running ffprobe [] exit status 1") ||
+			strings.Contains(err.Error(), "Could not create stream ID") ||
+			strings.Contains(err.Error(), "502 Bad Gateway") ||
+			(strings.Contains(err.Error(), "EOF") && strings.Contains(err.Error(), "error processing file")) {
+			err = errors.New("error processing file")
+		} else if strings.Contains(err.Error(), "MultipartUpload: upload multipart failed") {
+			err = errors.New("error saving file to internal storage")
+		} else if strings.Contains(err.Error(), "mp4io: parse error") {
+			err = errors.New("file format unsupported, must be MP4")
+		}
+	}
+
 	msg := event.AMQPMessage{
 		Exchange:   r.ExchangeName,
 		Key:        fmt.Sprintf("task.result.%s.%s", task.Type, task.ID),
--- a/task/runner.go
+++ b/task/runner.go
@@ -270,10 +270,10 @@ func humanizeError(err error) error {
 
 	if strings.Contains(errMsg, "unexpected eof") {
 		return errors.New("file download failed")
-	} else if strings.Contains(errMsg, "multipartUpload: upload multipart failed") {
+	} else if strings.Contains(errMsg, "multipartupload: upload multipart failed") {
 		return errors.New("internal error saving file to storage")
 	} else if strings.Contains(errMsg, "mp4io: parse error") {
-		return UnretriableError{errors.New("file format unsupported, must be MP4")}
+		return UnretriableError{errors.New("file format unsupported, must be a valid MP4")}
 	}
 
 	isProcessing := strings.Contains(errMsg, "error running ffprobe [] exit status 1") ||
@@ -283,6 +283,14 @@ func humanizeError(err error) error {
 
 	if isProcessing {
 		return errors.New("internal error processing file")
+	}
+
+	isTimeout := strings.Contains(errMsg, "context deadline exceeded") ||
+		strings.Contains(errMsg, "context canceled") ||
+		strings.Contains(errMsg, "context deadline exceeded")
+
+	if isTimeout {
+		return errors.New("execution timeout")
 	}
 
 	return err
--- a/task/runner.go
+++ b/task/runner.go
@@ -222,20 +222,7 @@ func (r *runner) getAssetAndOS(assetID string) (*api.Asset, drivers.OSSession, e
 func (r *runner) publishTaskResult(ctx context.Context, task data.TaskInfo, output *data.TaskOutput, err error) error {
 	resultCh := make(chan event.PublishResult, 1)
 
-	if err != nil {
-		if strings.Contains(err.Error(), "unexpected EOF") {
-			err = errors.New("file download failed")
-		} else if strings.Contains(err.Error(), "error running ffprobe [] exit status 1") ||
-			strings.Contains(err.Error(), "Could not create stream ID") ||
-			strings.Contains(err.Error(), "502 Bad Gateway") ||
-			(strings.Contains(err.Error(), "EOF") && strings.Contains(err.Error(), "error processing file")) {
-			err = errors.New("error processing file")
-		} else if strings.Contains(err.Error(), "MultipartUpload: upload multipart failed") {
-			err = errors.New("error saving file to internal storage")
-		} else if strings.Contains(err.Error(), "mp4io: parse error") {
-			err = errors.New("file format unsupported, must be MP4")
-		}
-	}
+	err = humanizeError(err)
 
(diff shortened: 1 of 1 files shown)


task/runner.go
```diff
@@ -221,6 +221,22 @@ func (r *runner) getAssetAndOS(assetID string) (*api.Asset, drivers.OSSession, e
...
+		} else if strings.Contains(err.Error(), "error running ffprobe [] exit status 1") ||
+			strings.Contains(err.Error(), "Could not create stream ID") ||
+			strings.Contains(err.Error(), "502 Bad Gateway") ||
+			(strings.Contains(err.Error(), "EOF") && strings.Contains(err.Error(), "error processing file")) {
+			err = errors.New("error processing file")
+		} else if strings.Contains(err.Error(), "MultipartUpload: upload multipart failed") {
+			err = errors.New("error saving file to internal storage")
+		} else if strings.Contains(err.Error(), "mp4io: parse error") {
+			err = errors.New("file format unsupported, must be MP4")
+		}
```
victorges, line 237: Could you create a helper for this logic so it doesn't distract from the main focus of this function? Then this could be as simple as
```suggestion
		err = humanizeError(err)
```
(you can even make the helper itself check if the `err != nil`, returning `nil` otherwise. Then here it's just 1 LOC 😌

victorges, line 237: As we chatted, need to wrap some errors (this cinluded) in an `UnretriableError` so it's not retried by the API.

task/runner.go
```diff
@@ -221,6 +221,22 @@ func (r *runner) getAssetAndOS(assetID string) (*api.Asset, drivers.OSSession, e
 
 func (r *runner) publishTaskResult(ctx context.Context, task data.TaskInfo, output *data.TaskOutput, err error) error {
 	resultCh := make(chan event.PublishResult, 1)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "unexpected EOF") {
```
victorges, line 226: Should probably get the `err.Error()` only once in the beginning, assign it to some variable and then just use it in these checks. 

It's not a good practice to call the same function over and over if you only need it's value once, since sometimes it can be just a waste in performance. In this case `Error()` functions are _generally_ very cheap, at worst only formatting some message, but it still leads to a repetition in the code that can be confusing, as if the response wouldn't always be the same from all those calls.

Also good to do a `strings.ToLower` on that error message so you don't need to worry about casing when checking for the error messages (tho make sure your string literals are all lowercase here): https://github.com/livepeer/go-api-client/blob/832e50712078b48755c17e203fa2f0433a98de93/api.go#L1116

task/runner.go
```diff
@@ -221,6 +221,22 @@ func (r *runner) getAssetAndOS(assetID string) (*api.Asset, drivers.OSSession, e
...
+			err = errors.New("file download failed")
+		} else if strings.Contains(err.Error(), "error running ffprobe [] exit status 1") ||
+			strings.Contains(err.Error(), "Could not create stream ID") ||
+			strings.Contains(err.Error(), "502 Bad Gateway") ||
+			(strings.Contains(err.Error(), "EOF") && strings.Contains(err.Error(), "error processing file")) {
```
victorges, line 231: Since this is the most verbose check here, I'd suggest you to leave it last (especially if you create a helper for humanizing these messages). I think it's a little easier to parse when there are only short and simple branches in the code and then 1 a little more complex one in the end. And if you have a helper, can even make early returns of the simple cases to make the complex part easier to digest e.g.
```go
func humanizeError(err error) error {
  if err == nil {
     return nil
  }
  errMsg := ...
  if errMsg ... {
     return errors.New(...)
  } else if
    // repeat a couple of these
 } 
   
 isProcessingError := strings.Contains(...) || ... && ...
 if isProcessingError {
    return errors.New(...)
 }
  return err
}
```

(actually, I think what would help the most there would be the helper variable, not only the fact that it is the last check. multi-line `if` conditions are very confusing IMO, even the default go indentation is weird. should try to avoid that whenever possible)

victorges (commented):
I think we should move all that logic to a helper function not to divert the focus of the `publishTaskResult` function.

task/runner.go
```diff
@@ -221,6 +221,22 @@ func (r *runner) getAssetAndOS(assetID string) (*api.Asset, drivers.OSSession, e
...
+			err = errors.New("error processing file")
+		} else if strings.Contains(err.Error(), "MultipartUpload: upload multipart failed") {
+			err = errors.New("error saving file to internal storage")
```
victorges, line 234: Hmm, just thought that we might wanna call these `internal error`s? WDYT?
```suggestion
			err = errors.New("internal error processing file")
		} else if strings.Contains(err.Error(), "MultipartUpload: upload multipart failed") {
			err = errors.New("internal error saving file to storage")
```

task/runner.go
gioelecerati: Much cleaner <3

task/runner.go
gioelecerati: Yeah, looks better! I was actually thinking about having a `error:` prefix like `error: Internal Error saving file to storage` for more readability, wdyt? For now I pushed your proposal

task/runner.go
```diff
@@ -258,4 +261,31 @@ func errorInfo(err error) *data.ErrorInfo {
...
+func humanizeError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	errMsg := strings.ToLower(err.Error())
+
+	if strings.Contains(errMsg, "unexpected eof") {
+		return errors.New("file download failed")
+	} else if strings.Contains(errMsg, "multipartUpload: upload multipart failed") {
```
victorges, line 273: got an uppercase here
```suggestion
	} else if strings.Contains(errMsg, "multipartupload: upload multipart failed") {
```

task/runner.go
```diff
@@ -258,4 +261,31 @@ func errorInfo(err error) *data.ErrorInfo {
...
+		return errors.New("internal error saving file to storage")
+	} else if strings.Contains(errMsg, "mp4io: parse error") {
+		return UnretriableError{errors.New("file format unsupported, must be MP4")}
```
victorges, line 276: WDYT of
```suggestion
		return UnretriableError{errors.New("file format unsupported, must be a valid MP4")}
```

task/runner.go
```diff
@@ -258,4 +261,31 @@ func errorInfo(err error) *data.ErrorInfo {
...
+	}
```
victorges, line 277: Could you add another check here for a new error we have?
```suggestion
	}
	
	isTimeout := strings.Contains(errMsg, "context deadline exceeded") ||
	    strings.Contains(errMsg, "context canceled") ||
	    strings.Contains(errMsg, "context deadline exceeded")
	 if isTimeout {
	     return errors.New("execution timeout")
	 }
```

victorges (approved):
Nice!

task/runner.go
victorges: Hmm, I think in the context where these messages will show up, it will already be clear that they correspond to an error message. So we don't need that prefix specifier. e.g.
```json
        "status": {
            "phase": "failed",
            "updatedAt": 1655333678345,
            "errorMessage": "error probing file: error running ffprobe [] exit status 1"
        },
```

Which with the current changes, should become
```json
        "status": {
            "phase": "failed",
            "updatedAt": 1655333678345,
            "errorMessage": "internal error processing file"
        },
```

WDYT? Still think that an `error:` prefix would be necessary?

gioelecerati: Definitely not necessary, just thought that uppercase errors in texts like `Internal Error` would provide more readability, but we can keep it like it is now! :)

task/runner.go
gioelecerati: Nice catch!

task/runner.go
gioelecerati: Makes a lot of sense, so we can directly skip questions about why the extension is correct but this error is returned

task/runner.go
victorges: Oh I thought you were only proposing an `error:` prefix to the messages. Did not understand the casing suggestion.

I like making these cased like sentences (first world Upper) instead of go errors (all lower) as well! Feel free to update these and just ignore the go vet warnings, it doesn't know what it's doing 😆

An alternative could also be to return a string from this function instead of an error. But you then would have to change the calling function as well, don't bother about it.

victorges (approved):
:shipit:

PR #65, opened 2022-09-14
task/prepare: Use asset playback recording OS ID for HLS

victorges:
For Storj integration.


victorges:
This should never go to prod. We will use catalyst VOD in which there will be only 1 OS per asset and we could use that.

PR #67, opened 2022-09-26
workflows: Update github workflow manifests.

hjpotter92:
- Add `dependabot` configuration
 - Merge `docker-amd64` jobs into a single manifest
 - Add CodeQL and tests workflow
 - Upload binaries to GCP


.github/workflows/docker-amd64.yaml
```diff
@@ -1,35 +0,0 @@
-# This publishes only the regular amd64 image to DockerHub. Only for a quicker
-# iteration since the cross-architecture build is really slow.
-name: AMD64 Docker image build
```
victorges, line 3: By removing this image, you need to update the values files in `livepeer-infra` to stop using the `amd64` tagged image as well..

Also, the only reason for this to exist was not to get a multi-platform builds, since I believe we already had that, but to get a build with a decent (but still high) iteration time, since the older multi-platform build took more than 10 minutes. The main reason for that time is installing `ffmpeg` which takes a couple of minutes on its own.

That's why I was also looking into Docker layer caching for the build here. Do you think it could work even for the multi-platform builds? I see that they are taking 8 minutes now, but that is still a pretty long time that I'll start doing something else and lose the train of focus 😆 

This is what I had found promising but didn't manage to implement cause we weren't using docker actions here yet:
https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching

Another alternative could be using the base image from `go-livepeer` that contains `ffmpeg` already installed. The one that stopped working as the base image on a `FROM` directive for reasons beyond my Docker knowledge. If we can bring that back here it'd probably be even better than the layer caching! (I think it was used from some docker action args now as "build cache" or sth?)

victorges (approved):
LGTM, but I think we should either keep the `amd64` builds or add layer caching to the full build or find a way to use that base image from `go-livepeer` (and stop using `amd64` on `livepeer-infr`).

IMO 8 minutes to get an image deployable on staging etc would be too long of an iteration time when hacking in this project. Today it is about half of that through that `amd64` quicker image instead.

Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY	.	.
 
 RUN	make "version=$version"
 
-FROM	debian:buster-slim
+FROM	livepeerci/build:latest
```
victorges, line 18: Should we remove the `apt install ffmpeg` below? Or is it quick when its already installed?

PR #76, opened 2022-10-26
task: Support ipfs:// URI imports

yondonfu:
This PR adds support for importing ipfs:// URLs by extracting the CID and trying to import the asset from a list of configured trusted gateways. The runner will try the next gateway in the list if the previous one did not successfully return the asset.

### DIFF
--- a/cmd/task-runner/task-runner.go
+++ b/cmd/task-runner/task-runner.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -53,6 +54,15 @@ func URLVarFlag(fs *flag.FlagSet, dest **url.URL, name, value, usage string) {
 	})
 }
 
+func StringsVarFlag(fs *flag.FlagSet, dest *[]string, name, value, usage string) {
+	*dest = strings.Split(value, ",")
+
+	fs.Func(name, usage, func(s string) error {
+		*dest = strings.Split(s, ",")
+		return nil
+	})
+}
+
 func parseFlags(build BuildFlags) cliFlags {
 	cli := cliFlags{}
 	fs := flag.NewFlagSet("livepeer-task-runner", flag.ExitOnError)
@@ -71,6 +81,7 @@ func parseFlags(build BuildFlags) cliFlags {
 	fs.StringVar(&cli.runnerOpts.PinataAccessToken, "pinata-access-token", "", "JWT access token for the Piñata API")
 	URLVarFlag(fs, &cli.runnerOpts.PlayerImmutableURL, "player-immutable-url", "ipfs://bafybeihcqgu4rmsrlkqvavkzsnu7h5n66jopckes6u5zrhs3kcffqvylge/", "Base URL for an immutable version of the Livepeer Player to be included in NFTs metadata")
 	URLVarFlag(fs, &cli.runnerOpts.PlayerExternalURL, "player-external-url", "https://lvpr.tv/", "Base URL for the updateable version of the Livepeer Player to be included in NFTs external URL")
+	StringsVarFlag(fs, &cli.runnerOpts.ImportIPFSGatewayURLs, "import-ipfs-gateway-urls", "Comma delimited ordered list of IPFS gateways (includes /ipfs/ suffix) to import assets from", "https://w3s.link/ipfs/,https://ipfs.io/ipfs/,https://cloudflare-ipfs.com/ipfs/")
 
 	// Server options
 	fs.StringVar(&cli.serverOpts.Host, "host", "localhost", "Hostname to bind to")
--- a/task/import.go
+++ b/task/import.go
@@ -17,6 +17,8 @@ import (
 	"github.com/livepeer/livepeer-data/pkg/data"
 )
 
+const IPFS_PREFIX = "ipfs://"
+
 type ImportTaskConfig struct {
 	// Ordered list of IPFS gateways (includes /ipfs/ suffix) to import assets from
 	ImportIPFSGatewayURLs []string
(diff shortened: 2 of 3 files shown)


task/import.go
```diff
@@ -110,7 +111,44 @@ func getFile(ctx context.Context, osSess drivers.OSSession, params api.UploadTas
...
+	// 	return getFileArweave(ctx, txID)
+	// }
+
+	return getFileWithUrl(ctx, params.URL)
+}
+
+func getFileIPFS(ctx context.Context, cid string) (name string, size uint64, content io.ReadCloser, err error) {
+	// Trusted public gateways
+	// These gateways will be tried in this order
+	gateways := []string{
```
yondonfu, line 133: Add ipfs.livepeer.studio

cmd/task-runner/task-runner.go
```diff
@@ -71,6 +81,7 @@ func parseFlags(build BuildFlags) cliFlags {
 	fs.StringVar(&cli.runnerOpts.PinataAccessToken, "pinata-access-token", "", "JWT access token for the Piñata API")
 	URLVarFlag(fs, &cli.runnerOpts.PlayerImmutableURL, "player-immutable-url", "ipfs://bafybeihcqgu4rmsrlkqvavkzsnu7h5n66jopckes6u5zrhs3kcffqvylge/", "Base URL for an immutable version of the Livepeer Player to be included in NFTs metadata")
 	URLVarFlag(fs, &cli.runnerOpts.PlayerExternalURL, "player-external-url", "https://lvpr.tv/", "Base URL for the updateable version of the Livepeer Player to be included in NFTs external URL")
+	StringsVarFlag(fs, &cli.runnerOpts.ImportIPFSGatewayURLs, "import-ipfs-gateway-urls", "Comma delimited ordered list of IPFS gateways (includes /ipfs/ suffix) to import assets from", "https://w3s.link/ipfs/,https://ipfs.io/ipfs/,https://cloudflare-ipfs.com/ipfs/")
```
0xcadams, line 84: This is a great approach to getting this in quickly. This may not resolve for a valid CID if the gateway doesn't have that content pinned. For example, if we don't include a gateway or if someone pins content on their own IPFS server. So a longer-term approach would probably be similar to Victor's suggestion of pinning the content on a pinning service (like Pinata) and allowing the service to perform the search/pin process, and we poll that API until ready, with some timeout (or if Pinata has its own timeout) for when we decide that the content cannot be found on the network.

0xcadams: This seems like a great first step though as long as we include as many IPFS providers as possible.

0xcadams (commented):
LGTM, will let Victor approve since he has more context

victorges (approved):
neat!

task/import.go
```diff
@@ -110,7 +116,36 @@ func getFile(ctx context.Context, osSess drivers.OSSession, params api.UploadTas
...
+	}
+
+	// TODO: Implement Arweave support
+	// arPrefix := "ar://"
+	// if strings.HasPrefix(params.URL, arPrefix) {
+	// 	txID := strings.TrimPrefix(params.URL, arPrefix)
+	// 	return getFileArweave(ctx, txID)
+	// }
+
+	return getFileWithUrl(ctx, params.URL)
```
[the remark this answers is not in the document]
yondonfu: Yeah! Can do a quick follow on PR for this.

cmd/task-runner/task-runner.go
yondonfu: I *think* the CID resolution problem could still exist even if we tried pinning the CID for out own gateway. IIUC if a gateway does not have a CID locally it will try to fetch it on the public network (DHT discovery + download the blocks from peers that have them). And if a pinning service pins a CID that it doesn't have locally it will also try to fetch it on the public network first before completing the pin. So, in both cases, if the fetching on the public network is slow or does not complete the CID wouldn't resolve.

Re-post from Discord:

> In the medium/long term, for IPFS, I think CAR validation is the ideal solution because we can download directly from a specified gateway that already has the CID which would be faster than re-pinning the CID in our own gateway which requires fetching the CID across the IPFS public network which can be slow. The problem right now is that CAR downloads are a relatively new feature in IPFS gateways and support across major gateways is still spotty - https://ipfs.github.io/public-gateway-checker/ shows that many public gateways still do not have full support for CAR downloads.

CAR validation is mentioned in more detail here https://docs.ipfs.tech/reference/http/gateway/#trusted-vs-trustless

task/import.go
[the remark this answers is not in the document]
yondonfu: Implemented in https://github.com/livepeer/task-runner/pull/77

0xcadams:
BTW doing more research on this, there are also IPFS providers which encourage the use of subdomains for the IPFS CID as opposed to the /ipfs/ path

https://bafybeib4euuddq6vbxt4uczd2o4omyjv64kctul6ss2bi7qwz6gsm36j4q.ipfs.dweb.link/

More details here:

https://consensys.net/diligence/blog/2021/06/ipfs-gateway-security/

cmd/task-runner/task-runner.go
victorges: I believe the behavior is like @yondonfu described as well! Using these public gateways, we will allow uploading any file available on the public IPFS network. Those gateways will resolve any files that can be found, only our own ipfs.livepeer.studio gateway that is explicitly configured for only returning files pinned in our account.

The immediate next step here will be to create an "open gateway" on our Piñata, which we can use here to access the IPFS network using our trusted provider. It will make things a little safer from any rate-limits we might get from these public gateways.

Both processes, finding a file by CID through an open gateway or pinning a CID in a known IPFS node (Piñata being our "node"), will require searching for the CID on the public network in the same way. The advantage of the latter, pinning in our own IPFS provider, would be that we'd guarantee that the file won't simply go away from IPFS. On the current approach, the owner can unpin the file from their provider and the file become unavailable on the IPFS network, but we'd keep supporting lookup using the "old" CID (which, in theory, is still valid, but we may virtually never be able to reproduce it without the DAG stored anywhere).

PR #77, opened 2022-10-26
task: Support ar:// URI imports

yondonfu:
Follow up to https://github.com/livepeer/task-runner/pull/76 to add support for ar:// URI imports by extracting the tx ID and importing the asset from the arweave.net gateway.


victorges (approved):
LGTM!

PR #82, opened 2022-11-03
task/export: Make 'file' the default NFT metadata

victorges:
It is more aligned with our new NFT and dStorage support strategy.

[moar ctx](https://discord.com/channels/423160867534929930/1009165020837466203/1037821695668531202)

### DIFF
--- a/task/export.go
+++ b/task/export.go
@@ -155,18 +155,15 @@ func nftMetadata(asset *api.Asset, videoCID string, template api.NFTMetadataTemp
 			"description":   fmt.Sprintf("Livepeer video from asset %q", asset.Name),
 			"image":         livepeerLogoUrl,
 			"animation_url": videoUrl,
-			"properties": map[string]interface{}{
-				"video": videoUrl,
-			},
 		}
 	case api.NFTMetadataTemplatePlayer:
 		return map[string]interface{}{
 			"name":          asset.Name,
 			"description":   fmt.Sprintf("Livepeer video from asset %q", asset.Name),
 			"image":         livepeerLogoUrl,
-			"animation_url": buildPlayerUrl(config.PlayerImmutableURL, asset.PlaybackID, true),
-			"external_url":  buildPlayerUrl(config.PlayerExternalURL, asset.PlaybackID, false),
-			// TODO: Consider migrating these to `attributes` instead.
+			"animation_url": buildPlayerUrl(config.PlayerImmutableURL, videoCID, true),
+			"external_url":  buildPlayerUrl(config.PlayerExternalURL, videoCID, false),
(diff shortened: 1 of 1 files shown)


task/export.go
```diff
@@ -152,27 +149,27 @@ func nftMetadata(asset *api.Asset, videoCID string, template api.NFTMetadataTemp
...
-	case api.NFTMetadataTemplateFile:
+	case api.NFTMetadataTemplatePlayer:
 		return map[string]interface{}{
 			"name":          asset.Name,
 			"description":   fmt.Sprintf("Livepeer video from asset %q", asset.Name),
 			"image":         livepeerLogoUrl,
-			"animation_url": videoUrl,
+			"animation_url": buildPlayerUrl(config.PlayerImmutableURL, asset.PlaybackID, true),
+			"external_url":  buildPlayerUrl(config.PlayerExternalURL, asset.PlaybackID, false),
+			// TODO: Consider migrating these to `attributes` instead.
```
0xcadams, line 169: Does this make sense to do as well, while we're moving to standardize the metadata? Would these would show up in OpenSea if we move them to `attributes`?

0xcadams (approved):
LGTM but not a code owner

task/export.go
victorges: So... the only real standard that exists is actually the `properties` field that we are using. The problem is it is on [ERC-1155](https://eips.ethereum.org/EIPS/eip-1155#metadata), not on 721 which only prescribes `name`, `description` and `image` fields.

I believe that while 1155 didn't exist, OpenSea ended up creating the `attributes` field on their own. Those are only documented [on their own docs](https://docs.opensea.io/docs/metadata-standards#metadata-structure), but it has also been followed by other marketplaces and become like a "social standard" among the NFT platforms. OpenSea does support the `properties` field and display them in the same way they would for attributes, but I believe they have some features on their own on top of special types of attributes (like displaying them differently). These additional features might not be available for `properties` tho and they don't even mention that in their docs, only `attributes`.

So IMO there's no obvious choice there. I'd prefer to stick by the decentralized standard and not an interfacce created by a single company. That interface has been adopted by other projects tho so it's not only from OpenSea anymore.

Going further a bit tho... I think that at least for the `file` template, we don't even need this property here. The file URL is already in the `animation_url`, so it's kind of redundant, and might become just another property that users would need to override on their own. Maybe we should remove `properties` altogether on that one, WDYT?

For the `player` template it's less obvious since the IPFS URL for the raw file is not anywhere else. But it's also less of priority right now, so we can leave it as is.

0xcadams: Yeah, removing for the `file` template and leaving it for the `player` template makes a lot of sense to me!

PR #88, opened 2022-11-19
config: Allow configuring runner params and human errors

### LINKED ISSUE #87: Hide catalyst errors in prod and expose pixel format error

victorges:
This is to allow configuring all the runner params including:
 - the number of concurrent tasks to execute, to allow consuming the queue more quickly without having to scale task runner
 - min/max task execution times
 - whether to humanize the error messages (we don't want to humanize in staging)

Also added humanization for the pixel format error from Catalyst as it's also needed for launch.

Fixes #87


cmd/task-runner/task-runner.go
```diff
@@ -88,6 +88,10 @@ func parseFlags(build BuildFlags) cliFlags {
 	fs.StringVar(&cli.runnerOpts.AMQPUri, "amqp-uri", "amqp://guest:guest@localhost:5672/livepeer", "URI for RabbitMQ server to consume from. Specified in the AMQP protocol")
 	fs.StringVar(&cli.runnerOpts.ExchangeName, "exchange-name", "lp_tasks", "Name of exchange where the task events will be published to")
 	fs.StringVar(&cli.runnerOpts.QueueName, "queue-name", "lp_runner_task_queue", "Name of task queue to consume from. If it doesn't exist a new queue will be created and bound to the API exchange")
+	fs.DurationVar(&cli.runnerOpts.MinTaskProcessingTime, "min-task-processing-time", task.DefaultMinTaskProcessingTime, "Minimum time that a task processing must take as rate-limiting strategy. If the task finishes earlier, the runner will wait for the remaining time before starting another task")
```
gioelecerati, line 91: Very cool

cmd/task-runner/task-runner.go
```diff
@@ -88,6 +88,10 @@ func parseFlags(build BuildFlags) cliFlags {
...
+	fs.DurationVar(&cli.runnerOpts.MaxTaskProcessingTime, "max-task-processing-time", task.DefaultMaxTaskProcessingTime, "Timeout for task processing. If the task is not completed within this time it will be marked as failed")
```
gioelecerati, line 92: WDYT of in the future specifying separate `MaxProcessingTime` at a task level? The task object could hold this value and be different depending on the type of task to execute, instead of having a global timeout specified on the runner.

task/runner.go
```diff
@@ -75,6 +81,15 @@ func NewRunner(opts RunnerOptions) Runner {
 	if opts.TaskHandlers == nil {
 		opts.TaskHandlers = defaultTasks
 	}
+	if opts.MinTaskProcessingTime == 0 {
```
gioelecerati, line 84: I am missing where `0` is set as default. I see that in the default we're setting `task.MinTaskProcessingTime`, which is not zero, so the only  time where this condition would be true is if we define `0` in the params?

cmd/task-runner/task-runner.go
```diff
@@ -88,6 +88,10 @@ func parseFlags(build BuildFlags) cliFlags {
...
+	fs.IntVar(&cli.runnerOpts.MaxConcurrentTasks, "max-concurrent-tasks", task.DefaultMaxConcurrentTasks, "Maximum number of tasks to run simultaneously")
```
gioelecerati, line 93: WDYT of making it a `UintVar` to avoid negative numbers being passed as a param? I see later we're only checking if the value is equal to `0`

task/runner.go
```diff
@@ -324,7 +341,9 @@ func (r *runner) scheduleTaskStep(ctx context.Context, taskID, step string, inpu
 }
 
 func (r *runner) publishTaskResult(ctx context.Context, task data.TaskInfo, output *data.TaskOutput, resultErr error) error {
-	resultErr = humanizeError(resultErr)
+	if r.HumanizeErrors {
```
gioelecerati, line 344: Just a curiosity, why is this flag needed? Don't we always want for the errors to be humanized?

task/runner.go
```diff
@@ -391,9 +410,16 @@ func humanizeError(err error) error {
 	if err == nil {
 		return nil
 	}
-
 	errMsg := strings.ToLower(err.Error())
 
+	var catErr CatalystError
+	if errors.As(err, &catErr) {
+		if strings.Contains(errMsg, "unsupported input pixel format") {
+			return errors.New("unsupported input pixel format, must be 'yuv420p' or 'yuvj420p'")
```
gioelecerati, line 418: I think the supported formats are defined into `supportedPixelFormats`, maybe we can log them here so when we increase our support we have a single point to edit

gioelecerati (approved):
LGTM with some comments and questions

cmd/task-runner/task-runner.go
victorges: Yeah that sounds like a good evolution

task/runner.go
victorges: It can be zero if one does not configure it when creating the options object, in which case we use the default here.

cmd/task-runner/task-runner.go
victorges: Sure

task/runner.go
victorges: Not on staging! Will save a few debugging round trips to have the raw error there

task/runner.go
victorges: It's a different config of supported stuff, and we don't read into that static list on our side when using Catalyst. So I just hardcoded them here while we don't get a better error from Catalyst like including the supported list or the file format

PR #89, opened 2022-11-22
task: Fix catalyst callback handling logic to avoid spill overs

victorges:
Today if a task is retried and the old catalyst job keeps sending callbacks to task-runner, we have
no way of differentiating those vs the actual callbacks we want to receive from the new job.

To fix this, added an `attemptId` query-string on the callback URLs so that task-runner can identify
the unique runs. This is not perfect, as if we start multiple jobs on the same task retry (e.g. event was
nacked and retried on RabbitMQ) we could still confuse the runs. 

It is already much better than before tho, as the direct-RabbitMQ retry is much rarer, and also easy to
evolve if we ever create some "internal storage" for the task runs to store intermediate data 
(like the Catalyst request ID). While we don't have that, this was the easiest I could think of.

### DIFF
--- a/api/handler.go
+++ b/api/handler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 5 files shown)


gioelecerati (approved):
LGTM

PR #91, opened 2022-11-30
clients: Add rate limit handling to catalyst client

victorges:
Simple logic, just waiting 15 seconds between each attempt. Should be fine since the request is also really cheap on catalyst side when it gets rate limited, and there will be at most 5 of these looping per task runner instance (3 instances)

My only concern now is that if we're operating at full capacity, these 5 executions will also block handling of any events from the currently running jobs. So we might stop checking if we're getting catalyst callbacks and handling the result callbacks (they become events as well).

Should consider reusing the "delayed task" abstraction used for checking catalyst callbacks for this as well, as a pre-step on the upload task execution in case it got rate limited.

Edit: Fixed the above and implemented an async back off through the delayed event.

### DIFF
--- a/clients/catalyst.go
+++ b/clients/catalyst.go
@@ -22,8 +22,8 @@ var (
 )
 
 const (
-	rateLimitRetryAfter = 15 * time.Second
-	maxAttempts         = 4
+	rateLimitRetryInitialDelay = 2 * time.Second
+	maxAttempts                = 4
 )
 
 type UploadVODRequest struct {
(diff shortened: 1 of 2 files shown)


thomshutt (approved):

clients/catalyst.go
```diff
@@ -70,15 +77,30 @@ func (c *catalyst) UploadVOD(ctx context.Context, upload UploadVODRequest) error
...
+			Body:        bytes.NewReader(body),
+			ContentType: "application/json",
+		}, &res)
+		glog.Infof("Catalyst upload VOD request: req=%q err=%q res=%q", body, err, string(res))
+
+		if !isTooManyRequestsErr(err) {
+			return err
+		}
+
+		if attempt >= maxAttempts {
```
thomshutt, line 94: So are we giving up if we get rate limited more than 3 times?

victorges: no! we will retry asynchronously instead through a delayed RabbitMQ event. Check the logic on `upload.go`

PR #100, opened 2022-12-12
Support transcode-file task

leszko:
Add a new task handler `TaskTranscodeFile`. It has the same workflow as `TaskUpload`, but it's unrelated to any Studio asset, therefore the logic is slightly different.

Related PRs:
- https://github.com/livepeer/go-api-client/pull/22
- https://github.com/livepeer/studio/pull/1548
----------

**Differences between `TaskTrancodeFile` and `TaskUpload`:**
1. `TaskUpload` can read input public video URL from UploadedObjectKey, but `TaskTranscodeFile` reads it only from the given URL.
2. `TaskTranscodeFile` allows to specify any bucket path as the output, while `TaskUpload` uses playbackId from the asset for the output path.
3. `TaskTrancodeFile` finalizes its  work when Catalyst Upload VOD pipeline is completed, while `TaskTranscodeFile` also verifies if the video is playable using playback URL.

This differences are expressed in the [handleUploadVODParams strategy](https://github.com/livepeer/task-runner/blob/333479463d471c5175fdced7f842de38a9a062a8/task/upload.go#L38).

--------

**Additional Comments:**
- The [related PR](https://github.com/livepeer/go-api-client/pull/22) in `livepeer/go-api-client` needs to get merged before this one
- I didn't add any unit tests, since there are currently almost no unit tests in `task-runner` 😱 I think we should definitely add them, but it would be separate activity / PR
- I think we should definitely do some task-runner refactoring / clean-up (e.g. remove deprecated tasks, create tasks as structs, rather than just functions, etc.), but also I think it'll be a separate PR


thomshutt (approved):
Very neat, I like that you were able to share most of the code with the Upload path.

Only thing I wasn't sure about was this bit in your PR description:

> TaskTranscodeFile also verifies if the video is playable using playback URL

Where's that happening?

leszko:
> > TaskTranscodeFile also verifies if the video is playable using playback URL
> 
> Where's that happening?

I might be wrong, but I think part of this code tries to download some files...

https://github.com/livepeer/task-runner/blob/ceaeecfe038fe9e01a6cdc57a51bc865ceac6673/task/upload.go#L292

task/upload.go
```diff
@@ -35,80 +35,126 @@ var (
...
+	finalize                 func(callback *clients.CatalystCallback) (*TaskHandlerOutput, error)
+	catalystPipelineStrategy pipeline.Strategy
+}
+
+func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
 	var (
-		ctx    = tctx.Context
-		step   = tctx.Step
-		params = *tctx.Task.Params.Upload
+		ctx  = p.tctx.Context
```
victorges, line 48: **nit:** Could have a local `tctx` variable here as well to avoid all the `p.tctx` changes below. Not a necessary change.

task/upload.go
```diff
@@ -117,24 +163,28 @@ func getFileUrl(os *api.ObjectStore, cfg ImportTaskConfig, params api.UploadTask
 		u.Path = path.Join(u.Path, params.UploadedObjectKey)
 		return u.String(), nil
 	}
-	if params.URL == "" {
+	return getFileUrl(cfg, params.URL)
+}
+
+func getFileUrl(cfg ImportTaskConfig, URL string) (string, error) {
```
victorges, line 169: **nit:** Local variables are generally lower case, so I think this would feel more natural.
```suggestion
func getFileUrl(cfg ImportTaskConfig, url string) (string, error) {
```

victorges (approved):
Clean implementation! LGTM, only some minor/optional nits :)

yondonfu:
@leszko 

> TaskTrancodeFile finalizes its work when Catalyst Upload VOD pipeline is completed, while TaskTranscodeFile also verifies if the video is playable using playback URL.

I think there might be a typo here and this should read as:

"TaskTrancodeFile finalizes its work when Catalyst Upload VOD pipeline is completed, while TaskUpload also verifies if the video is playable using playback URL."

Right?

leszko:
> > TaskTrancodeFile finalizes its work when Catalyst Upload VOD pipeline is completed, while TaskTranscodeFile also verifies if the video is playable using playback URL.
> 
> I think there might be a typo here and this should read as:
> 
> "TaskTrancodeFile finalizes its work when Catalyst Upload VOD pipeline is completed, while TaskUpload also verifies if the video is playable using playback URL."
> 
> Right?
Yes, you're right. Typing faster than thinking 🙃 

> Also, for clarification, I don't think "verifies if the video is playable using playback URL" is quite right. The code that you linked in [this comment](https://github.com/livepeer/task-runner/pull/100#issuecomment-1357297914) is responsible for handling some additional operations after results are returned by Catalyst - for example, the `complementCatalystPipeline()` function will [copy the source file from its URL](https://github.com/livepeer/task-runner/blob/ceaeecfe038fe9e01a6cdc57a51bc865ceac6673/task/upload.go#L324) into Studio's default bucket (i.e. GCP), [probe the source file](https://github.com/livepeer/task-runner/blob/ceaeecfe038fe9e01a6cdc57a51bc865ceac6673/task/upload.go#L371) among other things. These are operations that ideally Catalyst itself can handle going forward, but until then task-runner is "complementing" the Catalyst functionality within this function at the end of the `TaskUpload` flow. A better way to think about the conclusion of the `TaskUpload` flow may be that additional operations are executed to complete asset creation (i.e. gather probing metadata) and these additional operations are _not_ necessary for `TaskTranscodeFile` because the new task does not work with assets at all.

Yes, thanks for this clarification!

task/upload.go
leszko: Updated

task/upload.go
leszko: Updated

PR #101, opened 2022-12-13
Humanize some Catalyst errors

yondonfu:
Adds humanized versions of a few additional Catalyst errors:

Internal error type: 504 gateway timeouts during file import
Humanized error: file could not be imported from URL because it was not accessible

Internal error type: Unsupported video files for MediaConvert
Humanized error: invalid video file codec or container, check your input file against the input codec and container support matrix

Internal error type: Failure to probe or open files for MediaConvert
Humanized error: failed to probe or open file, check your input file against the input codec and container support matrix

These humanized errors should help with error filtering and categorization at the Studio API task level since these humanized errors can be stored as the error message for a failed task instead of the generic "internal error processing file"

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -475,6 +475,11 @@ func errorInfo(err error) *data.ErrorInfo {
 func humanizeCatalystError(err error) error {
 	errMsg := strings.ToLower(err.Error())
 
+	// General errors
+	if strings.Contains(errMsg, "import request") && strings.Contains(errMsg, "504 Gateway Timeout") {
+		return errors.New("file could not be imported from URL because it was not accessible")
+	}
+
 	// Livepeer pipeline errors
 	if strings.Contains(errMsg, "unsupported input pixel format") {
 		return errors.New("unsupported input pixel format, must be 'yuv420p' or 'yuvj420p'")
--- a/task/runner.go
+++ b/task/runner.go
@@ -471,6 +471,22 @@ func errorInfo(err error) *data.ErrorInfo {
 	return &data.ErrorInfo{Message: err.Error(), Unretriable: IsUnretriable(err)}
 }
 
+// Caller should check if err is a CatalystError first
+func humanizeCatalystError(err error) error {
+	errMsg := strings.ToLower(err.Error())
+
+	// Livepeer pipeline errors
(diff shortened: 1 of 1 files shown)


pglowacky:
+1 to humanized error messages

task/runner.go
```diff
@@ -471,6 +471,36 @@ func errorInfo(err error) *data.ErrorInfo {
...
+	// General errors
+	if strings.Contains(errMsg, "import request") && strings.Contains(errMsg, "504 Gateway Timeout") {
+		return errors.New("file could not be imported from URL because it was not accessible")
+	}
+
+	// MediaConvert pipeline errors
+	if strings.Contains(errMsg, "doesn't have video that the transcoder can consume") {
+		// TODO(yondonfu): Add link in this error message to a page with the input codec/container support matrix
+		return errors.New("invalid video file codec or container, check your input file against the input codec and container support matrix")
+	} else if strings.Contains(errMsg, "Failed probe/open") {
```
victorges, line 487: Notice that the error message is lowercased, so it will never match this. Needs to match against an all lowercase string instead.

Edit: realized that this came as a debt from the previous code that did these checks. I guess that has never worked 😯

task/runner.go
```diff
@@ -471,6 +471,36 @@ func errorInfo(err error) *data.ErrorInfo {
...
+		// TODO(yondonfu): Add link in this error message to a page with the input codec/container support matrix
+		return errors.New("failed to probe or open file, check your input file against the input codec and container support matrix")
+	}
+
+	// Livepeer pipeline errors
+	if strings.Contains(errMsg, "unsupported input pixel format") {
+		return errors.New("unsupported input pixel format, must be 'yuv420p' or 'yuvj420p'")
+	} else if strings.Contains(errMsg, "Unsupported video input") {
```
victorges, line 495: Same for casing

task/runner.go
```diff
@@ -471,6 +471,36 @@ func errorInfo(err error) *data.ErrorInfo {
...
+		return errors.New("unsupported file format")
+	} else if strings.Contains(errMsg, "ReadPacketData File read failed - end of file hit") {
```
victorges, line 497: Same

PR #104, opened 2022-12-14
task/runner: Handle extra import request errs when humanizing

### LINKED ISSUE #103: Improve handling of Catalyst error messages for inaccessible file when humanizing
[This check](https://github.com/livepeer/task-runner/blob/044f2766574bbc04255cdf8b28b1adbd06e7ffa6/task/runner.go#L479) needs to be able to handle the scenario where Catalyst retries and fails with a message like `Get "https://lens.infura-ipfs.io/ipfs/QmNjAMQ2uSmhoyjmcGsDj97VQLEMdTUsUt4fCxnmZAbJGk": GET https://lens.infura-ipfs.io/ipfs/QmNjAMQ2uSmhoyjmcGsDj97VQLEMdTUsUt4fCxnmZAbJGk giving up after 3 attempt(s)`

yondonfu:
Fixes https://github.com/livepeer/task-runner/issues/103

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -475,9 +475,17 @@ func errorInfo(err error) *data.ErrorInfo {
 func humanizeCatalystError(err error) error {
 	errMsg := strings.ToLower(err.Error())
 
+	fileNotAccessibleErrs := []string{
+		"504 Gateway Timeout",
+		"giving up after",
+	}
 	// General errors
-	if strings.Contains(errMsg, "import request") && strings.Contains(errMsg, "504 Gateway Timeout") {
-		return errors.New("file could not be imported from URL because it was not accessible")
+	if strings.Contains(errMsg, "import request") {
+		for _, e := range fileNotAccessibleErrs {
+			if strings.Contains(errMsg, e) {
+				return errors.New("file could not be imported from URL because it was not accessible")
+			}
+		}
 	}
 
 	// MediaConvert pipeline errors
--- a/task/runner.go
+++ b/task/runner.go
@@ -480,7 +480,9 @@ func humanizeCatalystError(err error) error {
 		"giving up after",
 	}
 	// General errors
-	if strings.Contains(errMsg, "import request") {
+	// TODO(yondonfu): This string matching is ugly and we should come up with a better less error-prone way to
+	// do this
+	if strings.Contains(errMsg, "download error") && strings.Contains(errMsg, "import request") {
 		for _, e := range fileNotAccessibleErrs {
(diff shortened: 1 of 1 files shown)


task/runner.go
```diff
@@ -475,9 +475,17 @@ func errorInfo(err error) *data.ErrorInfo {
 func humanizeCatalystError(err error) error {
 	errMsg := strings.ToLower(err.Error())
 
+	fileNotAccessibleErrs := []string{
+		"504 Gateway Timeout",
+		"giving up after",
```
thomshutt, line 480: I think we need a little more here, since there are a few places we do HTTP retries and this could happen for any of them

yondonfu: Do you think this will be a problem even we filter for both "download error" and "import request" first as is done in [7ae097f](https://github.com/livepeer/task-runner/pull/104/commits/7ae097f62e9b46458f4205f6d81f21bfd4dabde1)?

At the moment, I'm not sure if there is a better way that doesn't involve restructuring catalyst-api error propagation, but I added a TODO and can create an issue to consider restructuring error propagation.

thomshutt: You're right, I think that probably is enough - I'd missed the array was being used in combination with those

thomshutt (approved):

task/runner.go
yondonfu: Issue created https://github.com/livepeer/task-runner/issues/105

task/runner.go
```diff
@@ -475,9 +475,19 @@ func errorInfo(err error) *data.ErrorInfo {
 func humanizeCatalystError(err error) error {
 	errMsg := strings.ToLower(err.Error())
 
+	fileNotAccessibleErrs := []string{
+		"504 Gateway Timeout",
```
victorges, line 479: This needs to be lower case as well

task/runner.go
```diff
@@ -475,9 +475,19 @@ func errorInfo(err error) *data.ErrorInfo {
...
+		"giving up after",
+	}
 	// General errors
-	if strings.Contains(errMsg, "import request") && strings.Contains(errMsg, "504 Gateway Timeout") {
-		return errors.New("file could not be imported from URL because it was not accessible")
+	// TODO(yondonfu): This string matching is ugly and we should come up with a better less error-prone way to
```
victorges, line 483: Agreed. We should probably work with error codes like mediaconvert does (but may always have some string parsing fallback anyway, just hopefully won't be the main logic)

PR #110, opened 2022-12-14
doc: Add errors doc

yondonfu:
Add a doc to capture current error humanization information and the important external errors returned by task-runner for processing.

### DIFF
--- a/doc/processing-errors.md
+++ b/doc/processing-errors.md
@@ -1,0 +1,29 @@
+# Processing Errors
+
+task-runner by default will "humanize" any internal errors during task processing into external errors that are suitable to return to an end user. The `-humanize-errors` flag (and its corresponding env variable) can be used to disable error humanization (i.e. `-humanize-errors=false`).
+
+The code for humanizing errors can be found in the `humanizeError()` function.
+
+The current errors returned when error humanization is enabled:
+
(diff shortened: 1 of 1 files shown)


doc/processing-errors.md
```diff
@@ -0,0 +1,29 @@
+# Processing Errors
+
+task-runner by default will "humanize" any internal errors during task processing into external errors that are suitable to return to an end user. The `-humanize-errors` flag (and its corresponding env variable) can be used to disable error humanization (i.e. `-humanize-errors=false`).
```
leszko, line 3: Not strictly related to this PR, but in what scenarios do we use `-humanize-errors=true` and in which `-humanize-errors=false`?

I also thought `task-runner` is just an internal processing tool, so wonder why we would need 2 different behaviors.

leszko (approved):
Added one question, other than that LGTM

thomshutt (approved):

doc/processing-errors.md
yondonfu: `-humanize-errors=false` is set in our internal staging setup right now (see https://github.com/livepeer/livepeer-infra/blob/4a00610ea00a77ea070a340583cffa379784a76e/values/staging/task-runner.values.yaml#L6) to make it easier to sift through errors when testing.

PR #112, opened 2022-12-15
Fix 0/0 framerate handling for video tracks when probing

yondonfu:
This PR fixes a bug where the `parseFps()` function would return NaN for an input framerate of 0/0. The NaN value would be set as the value for a track's `FPS` field [here](https://github.com/livepeer/task-runner/blob/59c241c05d9541c34906a4e0786313cb26339d31/task/probe.go#L179). The track is nested in the larger [api.AssetSpec](https://github.com/livepeer/go-api-client/blob/c3675c55eed57d6d916f7db498af0956733e00ad/api.go#L329) struct which is marshalled in [saveMetadataFile()](https://github.com/livepeer/task-runner/blob/6853b65ac1ba193e59875e6a04383d73d0744972/task/import.go#L187) which is called during upload task execution after [task-runner receives a callback](https://github.com/livepeer/task-runner/blob/6853b65ac1ba193e59875e6a04383d73d0744972/task/upload.go#L321) from Catalyst in order to "complement" Catalyst which involves copying the source file into a bucket and also probing the file. The NaN value nested in the `api.AssetSpec` struct results in the following error in the call to `saveMetadataFile()`:

```
error marshaling file metadat: json: unsupported value: NaN
```

I observed this error for a file containing a mjpeg video track with the avg framerate set to 0/0.

### DIFF
--- a/task/probe_test.go
+++ b/task/probe_test.go
@@ -1,0 +1,51 @@
+package task
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseFps(t *testing.T) {
+	assert := assert.New(t)
+
+	// single value, invalid
+	_, err := parseFps("foobar")
+	assert.ErrorContains(err, "error parsing framerate:")
+
+	// single value
+	fps, err := parseFps("5")
+	assert.Nil(err)
+	assert.Equal(fps, 5.0)
+
+	// fraction, invalid
+	_, err = parseFps("foo/bar")
+	assert.ErrorContains(err, "error parsing framerate numerator:")
+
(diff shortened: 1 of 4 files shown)


thomshutt (approved):

task/probe_test.go
```diff
@@ -0,0 +1,56 @@
+package task
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseFps(t *testing.T) {
```
thomshutt, line 9: 😍

victorges, line 9: Nice! This is a necessary change in this project 🙊

Issue #109, opened 2022-12-14
(Catalyst) 404 Not Found when downloading file returned as internal error processing file


yondonfu:
https://github.com/livepeer/task-runner/pull/107/files#diff-33d20f5d5f41105e4334b7eb8ece1a7b748ed5e6f39acfab43ec28abfea3fed0R482 added a 404 Not Found string matching check. But, it doesn't seem to work because when humanizeErrors = true, the error returned by task-runner still seems to be internal error processing file. Related to Catalyst considering a 404 non-retryable when it reports a transcode status message?

Would be better if a 404 Not Found when downloading a file results in a file not accessible error returned by task-runner.

yondonfu:
Closed by #113

PR #113, opened 2022-12-15
task/runner: Fix bug with comparing lowercased errs

### LINKED ISSUE #111: Fix error humanization of Failed probe/open MC error
~Current logic does not capture the fact that the Failed probe/open can happen at upload time (current logic is focused on download time which probably doesn't make sense).~

The problem is that the error message is lowercased first so the string to compare against needs to be lowercased.

yondonfu:
Fixes https://github.com/livepeer/task-runner/issues/111 https://github.com/livepeer/task-runner/issues/109

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -26,6 +26,17 @@ const (
 	taskPublishTimeout           = 1 * time.Minute
 )
 
+// Humanized errors
+var (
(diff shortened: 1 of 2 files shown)


victorges (commented):
Ah I see you fixed those already! Sorry going chronologically over the emails.

I think the ones you added a todo here should be fixed as well. They definitely are not working today.

PR #122, opened 2023-01-05
Humanize no video codec error

red-0ne:
Translate "no video track found in file" errors to `errInvalidVideo`


victorges (approved):
LGTM but might be worth considering a custom error message in this case since it's so clear anyway. The other invalid video errors we have today we'd actually want to eventually parse the wrong thing (like "invalid video codec: x") then makes sense to keep it simple for now.

I'm fine with it as is anyway though. Might be worth checking if that's enough from the product perspective as well.

thomshutt (approved):

PR #124, opened 2023-01-10
Add metric for current task-runner version

thomshutt:


cmd/task-runner/task-runner.go
```diff
@@ -143,6 +144,9 @@ func Run(build BuildFlags) {
 	cli := parseFlags(build)
 	glog.Infof("Task runner starting... version=%q", build.Version)
 
+	// Fire a metric to track which version of task-runner we're running
+	metrics.Version.WithLabelValues("task-runner", build.Version).Inc()
```
thomshutt, line 148: @victorges open to moving this out of `Run()`, it was just less obvious where to put it within the `metrics` package

metrics/http.go
```diff
@@ -29,6 +29,13 @@ var (
 		},
 		[]string{"api"},
 	)
+	Version = Factory.NewCounterVec(
+		prometheus.CounterOpts{
```
hjpotter92, line 33: should it be counter or gauge? (few suggestions only stated gauge is better for one-off info). redis also uses gauge to export its version/deployment info as metrics

thomshutt: Maybe! I'll switch it, since I have no idea

cmd/task-runner/task-runner.go
victorges: Oh I thought there was a "high level" initialization function on the package, but maybe not in this project as the metric is also declared statically. If there's no such thing as a `metrics.Init` or `metrics.NewX`, this looks good

victorges (approved):

cmd/task-runner/task-runner.go
thomshutt: Yeah, we don't have any metrics initialisation function at the moment, so will leave it here for now

PR #125, opened 2023-01-17
Pass through dsotage urls

### LINKED ISSUE #1580: Pass through ipfs:// URLs to Catalyst API
**What's the problem? (required)**
Studio currently chooses an IPFS HTTP gateway for any uploads of `ipfs://` URLs and passes this into Catalyst API. This conversion should be handled natively in Catalyst API as per https://github.com/livepeer/catalyst-api/issues/141

**What's the current behavior? (required)**
Currently, conversion happens in https://github.com/livepeer/task-runner/blob/main/task/upload.go#L180

**Additional context (optional)**
This change should not happen until https://github.com/livepeer/catalyst-api/issues/141 is complete, tested and deployed.

red-0ne:
Since IPFS and Arweave urls are now handled by catalyst-api, there is no need to have this logic here and should just pass through the url.

Should close https://github.com/livepeer/studio/issues/1580

### DIFF
--- a/task/upload.go
+++ b/task/upload.go
@@ -8,7 +8,6 @@ import (
 	"net/url"
 	"path"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -135,7 +134,7 @@ func TaskUpload(tctx *TaskContext) (*TaskHandlerOutput, error) {
 
 func TaskTranscodeFile(tctx *TaskContext) (*TaskHandlerOutput, error) {
 	params := *tctx.Task.Params.TranscodeFile
-	inUrl, err := getFileUrl(tctx.ImportTaskConfig, params.Input.URL)
+	inUrl, err := getFileUrl(params.Input.URL)
 	if err != nil {
 		return nil, fmt.Errorf("error building file URL: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


task/upload.go
```diff
@@ -164,27 +163,14 @@ func getFileUrlForUploadTask(os *api.ObjectStore, cfg ImportTaskConfig, params a
 		u.Path = path.Join(u.Path, params.UploadedObjectKey)
 		return u.String(), nil
 	}
-	return getFileUrl(cfg, params.URL)
+	return getFileUrl(params.URL)
 }
 
-func getFileUrl(cfg ImportTaskConfig, url string) (string, error) {
+func getFileUrl(url string) (string, error) {
```
victorges, line 169: I wonder if we even need this function now. WDYT of moving this error checking onto `getFileUrlForUploadTask` and deleting this helper? In `TaskTranscode` we can let it pass through an error later, it is not an expected case at all anyway (the API would already block that).

task/upload.go
```diff
@@ -164,27 +163,14 @@ func getFileUrlForUploadTask(os *api.ObjectStore, cfg ImportTaskConfig, params a
 		u.Path = path.Join(u.Path, params.UploadedObjectKey)
 		return u.String(), nil
 	}
-	return getFileUrl(cfg, params.URL)
+	return getFileUrl(params.URL)
```
victorges, line 166: The caller here also has a `cfg` param which is now unused without this. Can you propagate the arg removal for cleanliness?

victorges (approved):
LGTM with some nits

task/upload.go
red-0ne: Was not sure if you intended to extend it in the future so I left it. Will remove it and refactor around :+1:

thomshutt (approved):

thomshutt:
Once this is tested and deployed, is there potentially task-runner config we can clean up (IPFS gateway list etc.) @victorges or is it still used elsewhere?

PR #126, opened 2023-01-19
Add profiles to Transcode API

leszko:


clients/catalyst.go
```diff
@@ -33,6 +34,7 @@ type UploadVODRequest struct {
 	Url              string            `json:"url"`
 	CallbackUrl      string            `json:"callback_url"`
 	OutputLocations  []OutputLocation  `json:"output_locations,omitempty"`
+	Profiles         []api.Profile     `json:"profiles,omitempty"`
```
leszko, line 37: This is maybe a little overuse, because the structure from the input API is used for the output API. So, to be 100% correct, we should define here exactly the same structure and rewrite all fields.

But being pragmatic, I think it should be ok, after all, we're passing through this param. Anyway, if you have an opinion, I'll change it and rewrite all fields.

thomshutt (approved):

victorges (approved):
LGTM

PR #128, opened 2023-01-23
humanize no pictures decoded error

mjh1:


victorges (approved):
LGTM!

For the future: At some point we should start handling mediaconvert error codes somehow as well instead of relying on error messages.

Perhaps we should support returning human errors directly from catalyst somehow, then we can treat all these errors with more accuracy over there

PR #129, opened 2023-01-24
runner: Make bad file errors unretriable

victorges:


task/runner.go
```diff
@@ -28,13 +28,13 @@ const (
 
 // Humanized errors
 var (
-	errFileInaccessible = errors.New("file could not be imported from URL because it was not accessible")
+	errFileInaccessible = UnretriableError{errors.New("file could not be imported from URL because it was not accessible")}
```
thomshutt, line 31: The other two make sense, but could this one potentially be temporary for e.g a flakey HTTP endpoint?

thomshutt (approved):

PR #132, opened 2023-02-01
task-runner: add mp4 output in catalyst request & callback

### LINKED ISSUE #130: Catalyst <-> Studio integration for Short form videos
After a sync with @emranemran , some work is needed to be done on Studio Task Runner and Catalyst for short form videos, specifically:

- add `AutoMP4s: bool` in `api/vod` in both the request studio & the catalyst-api struct
- Add `Mp4Outputs` to TranscodeStatusMessage from Catalyst to Studio
- Potentially, before populating the struct of TranscodeStatusMessage check on the storage if the mp4 is there (may require more time)
- `api/transcode` endpoint need to recognize the AutoMP4s

gioelecerati:
Fixes #130 
Reference https://github.com/livepeer/catalyst-api/issues/400


task/paths.go
```diff
@@ -16,6 +16,10 @@ func videoFileName(playbackID string) string {
 	return path.Join(playbackID, "video")
 }
 
+func mp4FileName(playbackID string) string {
+	return path.Join(playbackID, "mp4")
```
gioelecerati, line 20: @victorges if I understood correctly, this and the hls pathName should change to be a folder. Currently the hls pathname is `index.m3u8`, would you change it to `hls`?

task/upload.go
```diff
@@ -244,6 +245,19 @@ func processCatalystCallback(tctx *TaskContext, callback *clients.CatalystCallba
 			})
 		case OutputNameIPFSSourceMP4:
 			assetSpec.Storage.IPFS.CID = output.Manifest
+		case OutputNameAssetMP4:
+			if len(output.Videos) != 1 {
+				return nil, fmt.Errorf("unexpected number of videos in source MP4 output: %d", len(output.Videos))
+			}
+			video := output.Videos[0]
```
victorges, line 252: Need to loop over all videos.

task/paths.go
victorges: We only need to change the other file path if we are going with a single `output_location` in the request which is both HLS and MP4. If we are creating a separate `output_location` and parsing a separate `output` in the end, it doesn't really matter for now, and we can avoid changing any existing stuff.

If you do make any changes to use an array of a single `output_location`, the output folder should not be `hls`. It should be just `playbackId`, and the proper paths will be created internally by catalyst (`index.m3u8` or `hls/index.m3u8` for HLS, `mp4/*` for mp4s)

victorges (commented):
On the right path

task/upload.go
gioelecerati: Right! I haven't pushed, it's there now. Thanks for the catch

task/upload.go
```diff
@@ -244,6 +245,21 @@ func processCatalystCallback(tctx *TaskContext, callback *clients.CatalystCallba
...
+			if len(output.Videos) != 1 {
+				return nil, fmt.Errorf("unexpected number of videos in source MP4 output: %d", len(output.Videos))
+			}
+			for _, video := range output.Videos {
+				if video.Type != "mp4" {
+					return nil, fmt.Errorf("unexpected video type in source MP4 output: %s", output.Videos[0].Type)
+				}
+				videoFilePath = video.Location
+				assetSpec.Files = append(assetSpec.Files, api.AssetFile{
+					Type: "static_file",
```
victorges, line 258: WDYT of `"static_mp4` or even `static_transcoded_mp4` to be more specific? The other file we have is `catalyst_hls_manifest` in order to have a very clear meaning and processing logic. I think the more specific we are here the easier it will be to maintain without needing additional fields in this `asset.files` list, so no need for a generic name like "static file" IMO.

Additionally, we should add _some_ metadata about the file here. The MVP would be specifying the file size, but ideally we could have more video-specific info as well like resolution etc.

task/upload.go
```diff
@@ -407,6 +423,15 @@ func outputLocations(outURL string, relativePath string) ([]OutputName, []client
 				},
 			},
 		}
+	names, locations =
```
victorges, line 426: Actually, this should be under `assetOutputLocations`, not `outputLocations`, since we don't want this output in the case of the Transcode API.

task/upload.go
gioelecerati: I thought about it, I opted to left out `mp4` as in the future we may want to support other containers, specifically `webm` as, although is not widely supported, it's more modern and produce smaller files which could benefit the TTF & load time. However, since we are specifying `mp4` in other places, we can be specific also here and do something generic later and if needed, so I agree with this, changing it!

task/upload.go
gioelecerati: Good point! Moved it :)

task/upload.go
gioelecerati: I changed it as well on `studio` playback

task/upload.go
```diff
@@ -244,6 +245,21 @@ func processCatalystCallback(tctx *TaskContext, callback *clients.CatalystCallba
...
+				return nil, fmt.Errorf("unexpected number of videos in source MP4 output: %d", len(output.Videos))
+			}
+			for _, video := range output.Videos {
+				if video.Type != "mp4" {
+					return nil, fmt.Errorf("unexpected video type in source MP4 output: %s", output.Videos[0].Type)
+				}
+				videoFilePath = video.Location
+				assetSpec.Files = append(assetSpec.Files, api.AssetFile{
+					Type: "static_transcoded_mp4",
+					Path: videoFilePath,
```
victorges, line 259: I actually think we need to prepend this file path now with `mp4`, as we're adding that to the URL sent to catalyst for the output. Bonus if you can make it with a helper or something

task/upload.go
```diff
@@ -367,11 +383,21 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
...
 
 func assetOutputLocations(tctx *TaskContext) ([]OutputName, []clients.OutputLocation, error) {
-	outputNames, outputLocations, err := outputLocations(tctx.OutputOSObj.URL, tctx.OutputAsset.PlaybackID)
+	playbackId := tctx.OutputAsset.PlaybackID
+	outputNames, outputLocations, err := outputLocations(tctx.OutputOSObj.URL, playbackId)
 	if err != nil {
 		return nil, nil, err
 	}
-
+	mp4Url, _ := url.JoinPath(mp4FileName(playbackId))
```
victorges, line 391: You're using the wrong `url.JoinPath` function here, and is not sending a full URL to the `output_location` here. You must parse the object store URL like on line 420 🌿.

victorges (commented):
2 required changes

task/upload.go
gioelecerati: Yeah I was looking into that :eyes: It was importing `url` as the package, because outside of the outputLocations function I wasn't parsing the outUrl. Crazy to me that I can define a variable in a scope with the same name of a global package without go complaining, btw

task/upload.go
gioelecerati: Are you talking about `videoFilePath` having `mp4/` prepended here?

task/upload.go
victorges: Yeah that's really confusing haha
Fell for that a couple of times which is how I spotted it more easily here

task/upload.go
victorges: Yeah! Because the destionation we send to catalyst already contains that prefix in relation to the asset folder, so Catlayst should return a path relative to that, meaning we have to add the same prefix here. 

Bonus bonus if you can even use the same helper as for building output_location, but I think the logic is different lol

gioelecerati: Cool! Done with a separate helper hehe

victorges: Actually, what you should just use `extractOSUriFilePath` helper function! No need to repeat the `mp4` folder anywhere 😄

task/upload.go
```diff
@@ -244,6 +245,21 @@ func processCatalystCallback(tctx *TaskContext, callback *clients.CatalystCallba
...
+				return nil, fmt.Errorf("unexpected number of videos in source MP4 output: %d", len(output.Videos))
+			}
+			for _, video := range output.Videos {
+				if video.Type != "mp4" {
+					return nil, fmt.Errorf("unexpected video type in source MP4 output: %s", output.Videos[0].Type)
+				}
+				videoFilePath = video.Location
+				assetSpec.Files = append(assetSpec.Files, api.AssetFile{
+					Type: "static_transcoded_mp4",
+					Path: mp4FilePath(videoFilePath),
```
victorges, line 259: Sorry, I think catalyst actually returns the full path of the files.You should do the same as we do for `Manifest` instead and call `extractOSUriFilePath`

task/upload.go
gioelecerati: Mmmm not sure about how `extractOsUriFilePath` works, is this what you intended? https://github.com/livepeer/task-runner/pull/132/commits/82be9ae2ef0b6797234b422f382050abebe250e5

victorges: Yeah!

clients/catalyst.go
```diff
@@ -49,7 +49,7 @@ type OutputsRequest struct {
 	SourceMp4          bool `json:"source_mp4"`
 	SourceSegments     bool `json:"source_segments"`
 	TranscodedSegments bool `json:"transcoded_segments"`
-	AutoMP4s           bool `json:"auto_mp4s"`
+	AutoMP4            bool `json:"auto_mp4s"`
```
victorges, line 52: Forgot to update the JSON field

task/upload.go
```diff
@@ -467,6 +457,7 @@ func outputLocations(outURL string, relativePath string) ([]OutputName, []client
 				Outputs: &clients.OutputsRequest{
 					SourceSegments:     sourceSegments,
 					TranscodedSegments: true,
+					AutoMP4:            true,
```
victorges, line 460: This will set `AutoMP4` not only for `upload` but also for `transcode-file` tasks. I'd suggest receiving the `autoMP4` value as a boolean argument in this function, then setting it to true/false depending on where you call it from.

While at it, we could rename `assetOutputLocations` to `uploadTaskOutputLocations` to be clearer, wdyt?

victorges (commented):
cool

task/upload.go
gioelecerati: Oh right, makes sense.
And sure for the renaming!

task/paths.go
```diff
@@ -16,6 +16,10 @@ func videoFileName(playbackID string) string {
 	return path.Join(playbackID, "video")
 }
 
+func mp4FileName(playbackID string) string {
```
victorges, line 19: I think this is unused now?

task/upload.go
```diff
@@ -99,7 +101,7 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
 		}
 		glog.Infof("Processing upload vod catalyst callback. taskId=%s status=%q",
 			tctx.Task.ID, callback.Status)
-		if callback.Status != "success" {
+		if callback.Status != catalystClients.TranscodeStatusCompleted {
```
victorges, line 104: Uhm, I think this change was in `main`. Do you have merges in your branch or sth?

task/upload.go
```diff
@@ -228,8 +230,10 @@ func processCatalystCallback(tctx *TaskContext, callback *clients.CatalystCallba
 			return nil, fmt.Errorf("output type mismatch: %s != %s", output.Type, outReq.Type)
 		}
 		manifestPath, err := extractOSUriFilePath(output.Manifest, playbackID)
-		if err != nil {
-			return nil, fmt.Errorf("error extracting file path from output manifest: %w", err)
+		if outName != OutputNameAssetMP4 {
```
victorges, line 233: What is this check for? I don't think we have any output named that anymore, do we? Should probably remove the const and any references to it in the code.

task/upload.go
```diff
@@ -431,6 +457,7 @@ func outputLocations(outURL string, relativePath string) ([]OutputName, []client
 				Outputs: &clients.OutputsRequest{
 					SourceSegments:     sourceSegments,
 					TranscodedSegments: true,
+					AutoMP4:            autoMp4s,
```
victorges, line 460: super nit: We should probably rename this output (`OutputNameOSPlaylistHLS`) as it's not HLS anymore. I couldn't come up with a better name other than "OutputNameMain" which I don't like either, so not a priority. To keep in mind tho

victorges (approved):
LGTM!

PR #134, opened 2023-02-02
Web3 Storage

leszko:


task/upload.go
```diff
@@ -146,7 +147,14 @@ func TaskTranscodeFile(tctx *TaskContext) (*TaskHandlerOutput, error) {
 		},
 		finalize: func(callback *clients.CatalystCallback) (*TaskHandlerOutput, error) {
 			tctx.Progress.Set(1)
-			return &TaskHandlerOutput{}, nil
+			if callback.Outputs == nil || len(callback.Outputs) < 1 {
+				return nil, fmt.Errorf("invalid video outputs: %v", callback.Outputs)
+			}
+			return &TaskHandlerOutput{TaskOutput: &data.TaskOutput{
+				TranscodeFile: &data.TranscodeFileTaskOutput{
+					VideoFilePath: clients.RedactURL(callback.Outputs[0].Manifest),
```
victorges, line 155: Just to be sure, this will normally just be a public `ipfs://<cid>` URL right?

task/upload.go
```diff
@@ -395,14 +403,16 @@ func outputLocations(outURL string, relativePath string) ([]OutputName, []client
 	if err != nil {
 		return nil, nil, fmt.Errorf("error parsing object store URL: %w", err)
 	}
+	// source segments cannot be stored in web3.storage
+	sourceSegments := !strings.HasPrefix(outURL, "w3s")
```
victorges, line 407: Neat logic! A couple of questions:
 - Are we not storing HLS playlists in web3.storage at all, or only not the source segments rendition?
 - Does `catalyst-api` handle these settings correctly? I remember it got lost in its core pipeline, not sure if it's changed now.

victorges (approved):
LGTM! A few questions on the overall logic tho.

task/upload.go
leszko: Depending on the output storage system:
- For Storj => Redacted S3 URL
- For Web3.storage => Web3.storage gateway public URL

No sure, we should return it in the `ipfs://<cid>` format. I'll clarify it with Yondon.

task/upload.go
leszko: > * Are we not storing HLS playlists in web3.storage at all, or only not the source segments rendition?

We **do** store HLS playlist in manifest files, but **not** the source segments

> * Does `catalyst-api` handle these settings correctly? I remember it got lost in its core pipeline, not sure if it's changed now.

Yes, `catalyst-api` handles that correctly.

task/upload.go
leszko: Yeah, makes sense. Since it's web3.storage specific now, I think it's ok to return the URL of the w3s gateway for now.

victorges: Yeah agreed, especially since I just saw the corresponding code has already been merged. No need to bother about this now, let's keep as an idea for the future

thomshutt (approved):

PR #141, opened 2023-02-06
task: Allow decrypting input files with AESCBC

victorges:
This adds on top of https://github.com/livepeer/studio/pull/1645 to decrypt input files with
AES-256-CBC if a key was provided. 

Notice that this should ideally be done in Catalyst for a real implementation, but made this
quickly in order to create an experiment. The file is decrypted and copied to S3 instead of 
sent raw to Catalyst.

The experiment name is `vod-encrypted-input` and is checked by the API here: https://github.com/livepeer/livepeer-com/blob/579b4835ebfc15ca4700ac30b86a45159a4ddd1f/packages/api/src/controllers/asset.ts#L579

### DIFF
--- a/task/encryption/decrypt.go
+++ b/task/encryption/decrypt.go
@@ -1,0 +1,62 @@
+package encryption
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"encoding/hex"
+	"fmt"
+	"io"
+)
+
+func DecryptAES256CBCReader(reader io.ReadCloser, keyb16 string) (io.ReadCloser, error) {
+	key, err := hex.DecodeString(keyb16)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding base16 key: %w", err)
+	}
+
+	block, err := aes.NewCipher(key)
+	if err != nil {
+		return nil, fmt.Errorf("error creating cipher: %w", err)
+	}
+
+	iv := make([]byte, aes.BlockSize)
+	if _, err := reader.Read(iv); err != nil {
+		return nil, fmt.Errorf("error reading iv from input: %w", err)
+	}
+
+	stream := cipher.NewCBCDecrypter(block, iv)
+	pipeReader, pipeWriter := io.Pipe()
+
+	go func() {
+		defer reader.Close()
+		defer pipeWriter.Close()
+		defer func() {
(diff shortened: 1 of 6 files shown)


task/upload.go
```diff
@@ -300,20 +343,24 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
...
-			return nil, fmt.Errorf("error uploading file=%q to object store: %w", fullPath, err)
-		}
 		assetSpec.Files = append(assetSpec.Files, api.AssetFile{
 			Type: "source_file",
 			Path: toAssetRelativePath(playbackID, fullPath),
 		})
-		glog.Infof("Saved file=%s to url=%s", fullPath, fileUrl)
+
+		// in case of encrypted input, file will have been copied in the beginning
+		if params.Encryption.Key == "" {
```
gioelecerati, line 353: This is failing in all tests as params.Encryption is undefined, wrong version of lapi?

task/encryption/decrypt.go
```diff
@@ -0,0 +1,89 @@
...
+	for {
+		n, err := io.ReadFull(reader, buffer)
+		if n == 0 || err == io.EOF {
+			break
+		} else if err != nil && err != io.ErrUnexpectedEOF {
+			// unexpected EOF is returned when input ends before the buffer size
+			return err
+		} else if n%blockSize != 0 {
+			// this can only ever happen on the last chunk
+			return fmt.Errorf("input is not a multiple of AES block size. must be padded with PKCS#7")
```
gioelecerati, line 65: Do we want to try adding bytes at the end and try to pad it ourselves instead of returning this error?

task/upload.go
victorges: Yeah, probably was only working after merging with your `dev`

task/encryption/decrypt.go
victorges: I had that before before reading more about padding, but theoretically it's not in the spec to do that. If the encrypted content does not have full blocks it's necessarily malformed. I can add that logic back if you think it's better to be more resilient tho, we will detect the bad (decrypted) input when trying to probe the file anyway

victorges (commented):
self-review LGTM (after the prev fixes lol)

PR #147, opened 2023-02-10
humanize more catalyst errors

mjh1:


task/runner.go
```diff
@@ -526,7 +526,9 @@ func humanizeCatalystError(err error) error {
 		"readpacketdata file read failed - end of file hit",
 		"no video track found in file",
 		"no pictures decoded",
-		"error probing mp4 input file from s3",
+		"error running ffprobe [] exit status 1",
```
mjh1, line 529: @thomshutt do you think this is ok to assume invalid video for these errors (it's a different error for probe timeouts)

task/runner.go
```diff
@@ -526,7 +526,9 @@ func humanizeCatalystError(err error) error {
...
+		"zero bytes found for source",
+		"invalid framerate",
```
mjh1, line 531: @thomshutt are you happy for these to just be returned as "invalid video" too or should we create more specific errors

task/runner.go
victorges: Chiming in, I think this error is too generic for us to assume it is a problem with the input video. I think I've seen it happen on valid video files that we've got a problem ourselves with (like the MOOV box in the end thing). Since it's so generic, I think we could also get this error message from transient errors like network issues, etc, so it'd be bad to treat this as an "unretriable" error as well (which the `invalidVideoErrs` are.

WDYT, instead of treating this as a permanent "input error", creating a separate `error probing input file` human error, and having only this "system error" matching to that for now? Also making sure it is _not_ unretriable.

task/runner.go
thomshutt: I think "invalid video" is fine for now

task/runner.go
thomshutt: +1 to Victor's comments

mjh1: good point thanks

victorges (approved):
LGTM

PR #150, opened 2023-02-16
runner: humanize max resolution errors from mediaconvert pipeline

emranemran:


thomshutt (approved):

thomshutt:
This is a good humanization of the error, but should we be smarter in our code that generates the output profiles? Assuming this big resolution is because we're matching a high-resolution source, we could set maximums there and not match beyond a defined maximum (probably 4K)

mjh1 (approved):

victorges (approved):

emranemran:
> This is a good humanization of the error, but should we be smarter in our code that generates the output profiles? Assuming this big resolution is because we're matching a high-resolution source, we could set maximums there and not match beyond a defined maximum (probably 4K)

Yeah we should. Filed https://github.com/livepeer/catalyst-api/issues/468 to follow up. I'll go ahead and merge this so that we categorize these correctly for now until the Livepeer pipeline is fully in effect.

PR #157, opened 2023-02-27
task/runner: Add metrics on task latency and result

victorges:
We've created some alerts that use catalyst metrics to alert us. This is very limited
and has lead to a lot of noise in pagerduty since we have limited information in 
catalyst about how the errors should be handled, like when are they user errors.

With this, we'll be able to create a more e2e alert based on the actual error returned
by task runner, also filtering unretriable errors which are generally bad inputs which we
should not be alerted for.

Also created a metric for the latency executing tasks which will be good for monitoring.

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -587,12 +587,14 @@ func humanizeCatalystError(err error) error {
 
 	// Livepeer pipeline errors
 	if strings.Contains(errMsg, "unsupported input pixel format") {
-		return errors.New("unsupported input pixel format, must be 'yuv420p' or 'yuvj420p'")
-	} else if strings.Contains(errMsg, "Unsupported video input") {
-		// TODO(yondonfu): This check probably does not work because the error message will be lowercased
-		return errors.New("unsupported file format")
+		return UnretriableError{errors.New("unsupported input pixel format, must be 'yuv420p' or 'yuvj420p'")}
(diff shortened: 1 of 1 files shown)


task/runner.go
```diff
@@ -542,20 +596,22 @@ func humanizeCatalystError(err error) error {
...
 
 	// Livepeer pipeline errors
 	if strings.Contains(errMsg, "unsupported input pixel format") {
-		return errors.New("unsupported input pixel format, must be 'yuv420p' or 'yuvj420p'")
-	} else if strings.Contains(errMsg, "Unsupported video input") {
-		// TODO(yondonfu): This check probably does not work because the error message will be lowercased
-		return errors.New("unsupported file format")
+		return UnretriableError{errors.New("unsupported input pixel format, must be 'yuv420p' or 'yuvj420p'")}
+	} else if strings.Contains(errMsg, "unsupported video input") {
+		return UnretriableError{errors.New("unsupported file format")}
```
emranemran, line 609: Need this as part of the mc errors as well? 
https://github.com/livepeer/task-runner/pull/158/files

thomshutt (approved):

mjh1 (approved):

task/runner.go
victorges: Ah yeah! Removed this check as it's already caught on that list now

PR #160, opened 2023-02-28
task: added delete task handler

gioelecerati:


task/delete.go
```diff
@@ -0,0 +1,93 @@
...
+			for i := 0; i < maxRetries; i++ {
+				err = osSess.DeleteFile(ctx, filename)
+				if err == nil {
+					accumulator.Accumulate(1)
+					return nil
+				}
+				glog.Errorf("Error deleting file %v: %v (retrying...)", filename, err)
+				time.Sleep(retryInterval)
+			}
+			return fmt.Errorf("failed to delete file %v after %d retries", filename, maxRetries)
```
victorges, line 52: If you return an error the whole error group will cancel. You should always return `nil` and only log the errors instead (if you want to delete as much as possible before failing)

task/delete.go
```diff
@@ -0,0 +1,93 @@
...
+	files, err := osSess.ListFiles(ctx, directory, "/")
+
+	if err != nil {
+		glog.Errorf("Error listing files in directory %v", directory)
+	}
+
+	totalFiles := files.Files()
+
+	accumulator := NewAccumulator()
+	tctx.Progress.TrackCount(accumulator.Size, uint64(len(totalFiles)), 0)
```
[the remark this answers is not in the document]
victorges: Note: maybe the ideal long term solution might be to store the total size of the asset files in the asset object, when they are created or modified. Then here we could just measure the progress based on that total size. Not a change for this PR tho as it involves a lot other code paths.

task/delete.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if ipfs := asset.AssetSpec.Storage.IPFS; ipfs != nil {
+		err = UnpinFromIpfs(*tctx, ipfs.CID, "cid")
+		if err != nil {
+			glog.Errorf("Error unpinning from IPFS %v", ipfs.CID)
+		}
+		err = UnpinFromIpfs(*tctx, ipfs.NFTMetadata.CID, "nftMetadataCid")
+		if err != nil {
+			glog.Errorf("Error unpinning metadata from IPFS %v", ipfs.NFTMetadata.CID)
+		}
```
victorges, line 68: Actually, thinking further about it, if you never return errors, this task will always succeed and we will never even retry deleting the files. In order to simplify this whole logic (since you have some complexity missing like pagination), I'd suggest you to just return errors for now.

If you do want the failures to be as much partial successes as possible, you can accumulate the intermediate errors you get on a local slice here, and in the end you create a single error in case you found any errors during the process. I think that could be premature optimization for now tho, so I'd rather invest more in the additional retry logics you implemented here for file deletion, and could delete for IPFS unpinning as well. 

Can even create a generic helper function to retry another function a few times :D 
Geppetto gave me this:
```go
func retry(maxRetries int, f func() error) error {
    var err error

    for i := 0; i <= maxRetries; i++ {
        if err = f(); err == nil {
            return nil
        }
    }

    return fmt.Errorf("after %d attempts, last error: %s", maxRetries, err)
}
```

WDYT?

task/delete.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return &TaskHandlerOutput{
+		TaskOutput: &data.TaskOutput{},
```
victorges, line 72: Probably still create the `Delete` output field (empty object), otherwise we have some code in the API that can explode if the task type output is missing.

task/delete.go
```diff
@@ -0,0 +1,93 @@
...
+	}, nil
+}
+
+func UnpinFromIpfs(tctx TaskContext, cid string, filter string) error {
+	assets, _, err := tctx.lapi.ListAssets(api.ListOptions{
+		Filters: map[string]interface{}{
+			filter: cid,
```
victorges, line 79: Weird go syntax, it looks like the field is called `filter` lol

task/runner.go
```diff
@@ -35,14 +35,15 @@ var (
 	// without including extraneous error information from Catalyst
 	errInvalidVideo = UnretriableError{errors.New("invalid video file codec or container, check your input file against the input codec and container support matrix")}
 	// TODO(yondonfu): Add link in this error message to a page with the input codec/container support matrix
-	errProbe   = UnretriableError{errors.New("failed to probe or open file, check your input file against the input codec and container support matrix")}
+	errProbe = UnretriableError{errors.New("failed to probe or open file, check your input file against the input codec and container support matrix")}
```
victorges, line 38: I've fixed this on `main` already so you may get some conflicts

victorges (changes requested):
Requesting changes due to lack of pagination and the whole error handling story

PR #164, opened 2023-03-06
upload/ipfs: Grab video format from the right place

victorges:
I copied it from the export task where the asset spec is fully present, but during upload we
need to check the intermediate values like the result from catalyst itself or the local probing
(both on the `assetSpec` variable).

For that also moved the local probing to before the ipfs upload, or it still might've not been populated.


gioelecerati (approved):
LGTM

PR #170, opened 2023-03-21
task/upload: Allow asset IPFS metadata player template on creation 

victorges:
We had a legacy check on the export task to only allow assets with playback URLs to 
use player NFT metadata. This was because our very first assets didn't even have an HLS
playlist and just provided MP4 URLs.

For new assets creation, the only other place where `saveNFTMetadata` is called, this is not
necessary tho, as we will always create an HLS playback URL for the asset, or at least have it
playable on the player.

This fixes API-36

### DIFF
--- a/task/export.go
+++ b/task/export.go
@@ -98,6 +98,7 @@ func uploadFile(tctx *TaskContext, asset *api.Asset, content io.Reader) (*data.E
 			ipfs = clients.NewPinataClientAPIKey(p.APIKey, p.APISecret, extMetadata)
 		}
 	}
+
 	videoCID, metadata, err := ipfs.PinContent(tctx, "asset-"+asset.PlaybackID, contentType, content)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


task/export.go
```diff
@@ -98,11 +98,16 @@ func uploadFile(tctx *TaskContext, asset *api.Asset, content io.Reader) (*data.E
...
 		}
 	}
+
 	videoCID, metadata, err := ipfs.PinContent(tctx, "asset-"+asset.PlaybackID, contentType, content)
 	if err != nil {
 		return nil, err
 	}
 	template, nftMetadata := params.IPFS.NFTMetadataTemplate, params.IPFS.NFTMetadata
+	if template == api.NFTMetadataTemplatePlayer && asset.PlaybackURL == "" {
+		return nil, fmt.Errorf("cannot create player NFT for asset without playback URL")
```
0xcadams, line 108: So does this disable the `player` template on create?

victorges: No! For the `upload` task which is what we do on creation, we don't have that check. It should create a player metadata on creation!

0xcadams: Great!

0xcadams (approved):

PR #174, opened 2023-03-23
Humanize another invalid input error

mjh1:
We saw this error from mediaconvert for a 'bounce' variable resolution video: https://storage.googleapis.com/lp-us-catalyst-vod-com/hls/source/transfer/dfhhehfe

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -595,6 +595,7 @@ func humanizeCatalystError(err error) error {
 		"invalid framerate",
 		"maximum resolution is",
 		"unsupported video input",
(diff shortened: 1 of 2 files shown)


task/runner.go
```diff
@@ -595,6 +595,7 @@ func humanizeCatalystError(err error) error {
 		"invalid framerate",
 		"maximum resolution is",
 		"unsupported video input",
+		"scaler position rectangle is outside output frame",
```
victorges, line 598: fancy

victorges (approved):

PR #176, opened 2023-03-30
Add MP4 to Transcode API

leszko:
It also updates the Catalyst API VOD according to the design here: https://www.notion.so/livepeer/VOD-MP4-separate-from-HLS-34efd71b091a491ea4e785dda3bf9873

### DIFF
--- a/clients/catalyst.go
+++ b/clients/catalyst.go
@@ -45,10 +45,12 @@ type OutputLocation struct {
 }
 
 type OutputsRequest struct {
-	SourceMp4          bool `json:"source_mp4"`
-	SourceSegments     bool `json:"source_segments"`
-	TranscodedSegments bool `json:"transcoded_segments"`
-	AutoMP4            bool `json:"auto_mp4"`
+	SourceMp4 bool `json:"source_mp4"`
+	//SourceSegments     bool `json:"source_segments"`
+	//TranscodedSegments bool `json:"transcoded_segments"`
+	//AutoMP4            bool `json:"auto_mp4"`
+	HLS string `json:"hls"`
+	MP4 string `json:"mp4"`
 }
 
 type CatalystOptions struct {
--- a/clients/catalyst.go
+++ b/clients/catalyst.go
@@ -45,12 +45,9 @@ type OutputLocation struct {
 }
 
 type OutputsRequest struct {
-	SourceMp4 bool `json:"source_mp4"`
-	//SourceSegments     bool `json:"source_segments"`
-	//TranscodedSegments bool `json:"transcoded_segments"`
-	//AutoMP4            bool `json:"auto_mp4"`
-	HLS string `json:"hls"`
-	MP4 string `json:"mp4"`
+	SourceMp4 bool   `json:"source_mp4"`
+	HLS       string `json:"hls"`
+	MP4       string `json:"mp4"`
 }
 
 type CatalystOptions struct {
--- a/task/paths.go
+++ b/task/paths.go
@@ -16,7 +16,7 @@ func videoFileName(playbackID string) string {
 	return path.Join(playbackID, "video")
 }
 
-func hlsRootPlaylistFileName(playbackID string) string {
+func outputLocation(playbackID string) string {
 	return path.Join(playbackID)
 }
 
--- a/task/paths.go
+++ b/task/paths.go
@@ -16,10 +16,6 @@ func videoFileName(playbackID string) string {
 	return path.Join(playbackID, "video")
 }
 
-func outputLocation(playbackID string) string {
-	return path.Join(playbackID)
-}
-
 func metadataFileName(playbackID string) string {
 	return path.Join(playbackID, "video.json")
 }
--- a/task/paths.go
+++ b/task/paths.go
@@ -17,7 +17,7 @@ func videoFileName(playbackID string) string {
 }
 
 func hlsRootPlaylistFileName(playbackID string) string {
-	return path.Join(playbackID, "index.m3u8")
+	return path.Join(playbackID)
 }
 
 func metadataFileName(playbackID string) string {
--- a/task/upload.go
+++ b/task/upload.go
@@ -199,7 +199,9 @@ func toTranscodeFileTaskOutput(outputs []video.OutputVideo) (data.TranscodeFileT
 		return res, err
 	}
 	res.BaseUrl = bu
-	res.Hls = &data.TranscodeFileTaskOutputPath{Path: p}
+	if len(o.Videos) > 0 {
+		res.Hls = &data.TranscodeFileTaskOutputPath{Path: p}
+	}
 
 	for _, m := range o.MP4Outputs {
 		_, p, err := parseUrlToBaseAndPath(m.Location)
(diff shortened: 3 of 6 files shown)


task/paths.go
```diff
@@ -16,8 +16,8 @@ func videoFileName(playbackID string) string {
 	return path.Join(playbackID, "video")
 }
 
-func hlsRootPlaylistFileName(playbackID string) string {
-	return path.Join(playbackID, "index.m3u8")
+func outputLocation(playbackID string) string {
+	return path.Join(playbackID)
```
thomshutt, line 20: Is this method doing anything now?

leszko: hehe, right, removed.

task/upload.go
```diff
@@ -517,7 +527,7 @@ func removeCredentials(metadata *clients.CatalystCallback) *clients.CatalystCall
 func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.OutputLocation, error) {
 	playbackId := tctx.OutputAsset.PlaybackID
 	outURL := tctx.OutputOSObj.URL
-	outputNames, outputLocations, err := outputLocations(outURL, playbackId, true)
+	outputNames, outputLocations, err := outputLocations(outURL, "enabled", playbackId, "only_short", playbackId)
```
thomshutt, line 530: Could we have these as constants?

task/upload.go
```diff
@@ -539,23 +549,26 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
 	return outputNames, outputLocations, nil
 }
 
-func outputLocations(outURL string, relativePath string, autoMp4s bool) ([]OutputName, []clients.OutputLocation, error) {
+func outputLocations(outURL string, hls string, hlsRelPath string, mp4 string, mp4RelPath string) ([]OutputName, []clients.OutputLocation, error) {
```
thomshutt, line 552: You didn't introduce the method, so looks like it was already untested but given we have a fairly discrete set of inputs and outputs it should be pretty unit testable - would you mind adding a test in here?

thomshutt (approved):
Other than my comments this LGTM - I don't think either of them is a blocker, so your call on addressing them

clients/catalyst.go
```diff
@@ -45,10 +45,9 @@ type OutputLocation struct {
 }
 
 type OutputsRequest struct {
-	SourceMp4          bool `json:"source_mp4"`
-	SourceSegments     bool `json:"source_segments"`
-	TranscodedSegments bool `json:"transcoded_segments"`
-	AutoMP4            bool `json:"auto_mp4"`
+	SourceMp4 bool   `json:"source_mp4"`
+	HLS       string `json:"hls"`
+	MP4       string `json:"mp4"`
```
victorges, line 50: Wow is this already supported?

task/upload.go
victorges, line 552: **niiiit:** can drop a few characters here lol
```suggestion
func outputLocations(outURL, hls, hlsRelPath, mp4, mp4RelPath string) ([]OutputName, []clients.OutputLocation, error) {
```

task/upload.go
```diff
@@ -539,23 +549,26 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
...
 		[]clients.OutputLocation{
 			{
 				Type: "object_store",
-				URL:  url.JoinPath(hlsRootPlaylistFileName(relativePath)).String(),
+				URL:  url.JoinPath(hlsRelPath).String(),
+				Outputs: &clients.OutputsRequest{
+					HLS: hls,
+				},
+			},
+			{
```
victorges, line 567: Just double-checking: Does catalyst support multiple `output_locations` provided on the request now? Previously it only read the first one and checked the outputs

task/upload.go
```diff
@@ -539,23 +549,26 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
...
-func outputLocations(outURL string, relativePath string, autoMp4s bool) ([]OutputName, []clients.OutputLocation, error) {
+func outputLocations(outURL string, hls string, hlsRelPath string, mp4 string, mp4RelPath string) ([]OutputName, []clients.OutputLocation, error) {
 	url, err := url.Parse(outURL)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error parsing object store URL: %w", err)
 	}
-	// source segments cannot be stored in web3.storage
-	sourceSegments := !strings.HasPrefix(outURL, "w3s")
 	names, locations :=
 		[]OutputName{OutputNameOSPlaylistHLS},
```
victorges, line 558: You have to add another value to this slice here. Let me share a bit more context of why this `outputLocations` function exists (my bad for not having documented it originally, this was all done in a rush): 
 - The original Catalyst API provided no way to associated outputs from the requested output_locations
 - It actually ended up supporting only 1 location and 1 output in the first version, but it received and returned an array
 - There would also be other output types that would be incompatible with sending everything on a single output_location: e.g. saving to IPFS, which would have a completely different URL with pinata creds instead of S3

So we created this function here, which would create outputs in a deterministic order and then we did a logic on `processCatalystCallback` that assumes Catalyst also returns the `outputs` on the exact order as they were requested [1]. That's the whole reason for having those `names` here which are also returned and then processed in order [on the big switch in the process result function](https://github.com/livepeer/task-runner/blob/b112d7458c808c7d01f54afdab2f06ccae706368/task/upload.go#L332).

[1] (but we don't actually use the same request data from the task params, we regenerate the outputs which is error prone in case of deploys)

I'm not sure if either of those constraints are still true. Maybe we can already associate specific `output_locations` with their corresponding `outputs` now. Or maybe the output data became different enough that we can actually tell which is which without trying to associate.

Either way, I never liked this logic, so feel free to refactor it completely. For now though, since you are adding a new output location, you might need to add a new output name and make sure that `processCatalystCallback` is able to process it consistently. In the long run it'd be good to have a safer way to associate the 2 things though, like an output_location id, or maybe the outputs just embed the entire output_location requests, idk!

clients/catalyst.go
leszko: Yeah, in the other PR: https://github.com/livepeer/catalyst-api/pull/546

task/upload.go
leszko: Ok, I get it now. It's very confusing, for now, I'll add another output here, but we should think about tidying it up a little bit.

task/upload.go
leszko: Sure, moved to constants

task/upload.go
leszko: Yes, it will support it after this PR is merged: https://github.com/livepeer/catalyst-api/pull/546

task/upload.go
leszko: Added unit tests.

mjh1 (approved):

task/upload.go
victorges: Yeah agreed. We need some reliable way to map outputs to the requests somehow

PR #197, opened 2023-05-18
upload: added encryption to upload task params

gioelecerati:


task/upload.go
```diff
@@ -76,6 +76,14 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
 		if err != nil {
 			return nil, err
 		}
+		var encryption *clients.EncryptionPayload
+		params := *tctx.Task.Params.Upload
+
+		if params.Encryption.EncryptedKey != "" {
+			encryption = &clients.EncryptionPayload{
+				EncryptedKey: params.Encryption.EncryptedKey,
```
victorges, line 84: You need to check whether the upload params are not nil. I've made that mistake with that old experiment and it would break transcode API (should we support encryption in transcode API?)

Also while at it, do you mind finding where the old Encryption.Key was used and possibly deleting that code? Or can be done as a follow-up (there are other repos to update)

victorges (approved):

task/upload.go
gioelecerati: Of course! Yeah, the old code needs to be deleted both here and on studio, currently on studio it's just commented out on the experiment check

cmd/task-runner/task-runner.go
```diff
@@ -156,6 +159,18 @@ func Run(build BuildFlags) {
 	runner := task.NewRunner(cli.runnerOpts)
 	ctx := contextUntilSignal(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	eg, ctx := errgroup.WithContext(ctx)
+
+	if cli.runnerOpts.VodDecryptPrivateKey != "" && cli.runnerOpts.VodDecryptPublicKey != "" {
```
victorges, line 163: Can you add a // TODO everywhere you can saying that this should be moved to Catalyst-only? It hurts that we need this hack, but I agree with the approach not to need a huge refactor.

task/upload.go
```diff
@@ -62,20 +62,20 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
...
-		// TODO: Move this input decryption logic to catalyst
-		if uploadParams := tctx.Params.Upload; uploadParams != nil {
-			var err error
-			inUrl, err = decryptInputFile(tctx, inUrl, *uploadParams)
-			if err != nil {
-				return nil, fmt.Errorf("error decrypting input file: %w", err)
-			}
-		}
 		fallthrough
 	case "rateLimitBackoff":
```
victorges, line 66: Delete this separate `""` case since we don't need any custom logic anymore. This should be
```go
case "", "rateLimitBackoff":
```

task/upload.go
```diff
@@ -62,20 +62,20 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
...
 			return nil, err
 		}
+		var encryption *clients.EncryptionPayload
+		params := tctx.Task.Params.Upload
+
+		if params != nil && params.Encryption.EncryptedKey != "" {
+			encryption = &clients.EncryptionPayload{
+				EncryptedKey: params.Encryption.EncryptedKey,
+			}
+		}
```
victorges, line 78: Can you move this to `handleUploadVODParams` as I suggested? This function should not need to know or care what is the task type of the caller. It was made exactly to abstract the common logic between all of them (upload or transcode-file). 

So rather of re-checking the same conditions from which its args are constructed, it should have its arguments well defined via its input struct arg.

task/upload.go
```diff
@@ -553,9 +522,11 @@ func removeCredentials(metadata *clients.CatalystCallback) *clients.CatalystCall
 func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.OutputLocation, error) {
 	playbackId := tctx.OutputAsset.PlaybackID
 	outURL := tctx.OutputOSObj.URL
-	var mp4 = OUTPUT_ONLY_SHORT
+	var mp4 string
 	if tctx.OutputAsset.Source.Type == "recording" {
 		mp4 = OUTPUT_ENABLED
+	} else {
+		mp4 = OUTPUT_ONLY_SHORT
```
victorges, line 529: Why this change?

webcrypto/decryption.go
victorges, line 1: Didnt check this file just assumed it was the same.

victorges (approved):

PR #199, opened 2023-06-12
Update catalyst for new transcode status field

mjh1:
Updating catalyst to bring in new source playback field, that way we can update catalyst to start sending that without breaking the parsing on the task-runner side.


victorges (approved):

victorges:
> that way we can update catalyst to start sending that without breaking the parsing on the task-runner side.

FTR I don't think parsing would break due to a new field! Only if you changed the type of an existing field e.g. an array becomes an object. Good to be safe anyway.

PR #202, opened 2023-06-22
Add export-data task

leszko:
Add export-data task.

Related studio PR: https://github.com/livepeer/studio/pull/1775


task/export.go
```diff
@@ -61,24 +63,75 @@ func TaskExport(tctx *TaskContext) (*TaskHandlerOutput, error) {
...
+	name := "asset-" + asset.PlaybackID
+	contentType := "video/" + asset.VideoSpec.Format
+	cid, internalMetadata, err := uploadFile(tctx, customParams, ipfsParams, name, content, contentType)
+	if err != nil {
+		return nil, err
+	}
+	nftCid, err := uploadNftMetadata(tctx, ipfsParams, cid)
+	if err != nil {
+		return nil, err
+	}
```
victorges, line 76: We probably need to check if `ipfsParams != nil` before calling this, just to still support that custom export one (not critical, not sure if anyone has ever used that tbh)

task/export.go
```diff
@@ -61,24 +63,75 @@ func TaskExport(tctx *TaskContext) (*TaskHandlerOutput, error) {
...
+				IPFS: &data.IPFSExportInfo{
+					VideoFileCID:   cid,
+					NFTMetadataCID: nftCid,
+				},
+			},
+		},
+	}, nil
+}
+
+func TaskExportData(tctx *TaskContext) (*TaskHandlerOutput, error) {
```
victorges, line 90: Would you mind moving this to a separate `export_data` file? Since we're not sharing the same entrypoint, mostly the upload file helpers, I think it couldbe simpler to keep them in separate files.

task/export.go
```diff
@@ -61,24 +63,75 @@ func TaskExport(tctx *TaskContext) (*TaskHandlerOutput, error) {
...
+	var (
+		params       = tctx.Params.ExportData
+		customParams = params.Custom
+		ipfsParams   = params.IPFS
+	)
+	content := strings.NewReader(params.Content)
+	name := fmt.Sprintf("%s-%s", params.Type, params.ID)
+	contentType := "application/json"
```
victorges, line 98: If we only support JSON, might want to make the `params.Content` a `json.RawMessage` instead. We'd avoid encoding the JSON as a string as well as validate that the data is actually a JSON (while the type itself is just an alias for `[]byte`, can use `bytes.NewReader` as well)

task/import.go
```diff
@@ -97,16 +97,16 @@ func TaskImport(tctx *TaskContext) (*TaskHandlerOutput, error) {
 
 func getFile(ctx context.Context, osSess drivers.OSSession, cfg ImportTaskConfig, params api.UploadTaskParams, vodDecryptPrivateKey string) (name string, size uint64, content io.ReadCloser, err error) {
 	name, size, content, err = getFileRaw(ctx, osSess, cfg, params)
-	if err != nil || params.Encryption.EncryptedKey == "" {
+	if err != nil || params.Encryption.Key == "" {
```
victorges, line 100: We actually can't change this! The `Key` was the old field and `EncryptedKey` is the current one.

@gioelecerati found out you ended up never merging this: https://github.com/livepeer/go-api-client/pull/41
Probably only worked cause we released `task-runner` with a ref to the PR branch.

victorges (approved):
LGTM, only need to revert the `EncryptedKey` change (check comment)

task/import.go
leszko: @gioelecerati I've just merged your PR: https://github.com/livepeer/go-api-client/pull/41 :)

leszko: Reverted EncryptedKey changes in b00095719f95be725960650b9aba996feb8fa425

task/export.go
leszko: Right, good catch. Updated in 9f5ed61c0588d6a81c41ac5524b176c9b48cbf95

task/export.go
leszko: No problem. Moved in 74b5325077d29a5bce09019e549e50453c201373

task/export.go
leszko: Updated in a6b5ea2c88768cca6a3e3aaca2f861ea68fec877

PR #203, opened 2023-06-29
Single download pipeline

mjh1:
Initial step towards avoiding task runner needing to download the source file. We check if the file already exists (from catalyst) and skip making the copy at task runner if so. Both cases have been tested by using a local version of catalyst-api with the source copy happening and without.

Also removing the metadata file which we deemed is not needed.

### DIFF
--- a/task/upload.go
+++ b/task/upload.go
@@ -412,14 +412,31 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
 		osSess               = tctx.outputOS // Upload deals with outputOS only (URL -> ObjectStorage)
 		inFile               = params.URL
 		vodDecryptPrivateKey = tctx.VodDecryptPrivateKey
+		contents             io.ReadCloser
+		size                 uint64
+		filename             string
+		catalystCopiedSource = false
 	)
 	if isHLSFile(inFile) {
 		return &data.UploadTaskOutput{AssetSpec: assetSpec}, nil
 	}
-	filename, size, contents, err := getFile(tctx, osSess, tctx.ImportTaskConfig, params, vodDecryptPrivateKey)
-	if err != nil {
-		return nil, fmt.Errorf("error getting source file: %w", err)
+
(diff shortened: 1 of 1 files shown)


leszko (commented):
Added one comment. Other than that, I have 2 concerns (discussed in Discord):
1. Do we need this PR if we plan to remove copying from task-runner at all?
2. The output asset has changed. Isn't that an issue?

task/upload.go
```diff
@@ -503,15 +517,6 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
 		assetSpec.Storage.IPFS = &ipfs
 	}
 
-	_, metadataPath, err := saveMetadataFile(tctx, tctx.outputOS, tctx.OutputAsset.PlaybackID, metadata)
```
leszko, line 506: What is this metadata, don't we need it anymore?

task/upload.go
```diff
@@ -457,22 +474,19 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
...
-		metadata, err = Probe(tctx, tctx.OutputAsset.ID, filename, input, false)
+		metadata, err := Probe(tctx, tctx.OutputAsset.ID, filename, input, false)
 		if err != nil {
 			return nil, err
 		}
 		probed := metadata.AssetSpec
 		assetSpec.Hash, assetSpec.Size, assetSpec.VideoSpec = probed.Hash, probed.Size, probed.VideoSpec
 	}
 
-	metadata.AssetSpec, metadata.CatalystResult = &assetSpec, removeCredentials(callback)
```
leszko, line 474: Don't we need it anymore?

leszko (approved):
LGTM

task/upload.go
mjh1: This is the file I brought up on discord, seems like we're good to remove it. I'll wait for @victorges to check the PR though before merging.

PR #204, opened 2023-07-05
Enable catalyst source copying

mjh1:
- This enables source copying at catalyst for upload tasks (and not transcode-file) by setting the SourceMp4 flag on the catalyst request.
- Removed unused function `removeCredentials`

### DIFF
--- a/task/upload.go
+++ b/task/upload.go
@@ -457,7 +457,7 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
 
 	if !catalystCopiedSource {
 		// in case of encrypted input, file will have been copied in the beginning
-		if params.Encryption.EncryptedKey == "" {
+		if !isEncryptionEnabled(params) {
 			input, err := readLocalFile(0.95)
 			if err != nil {
 				return nil, err
@@ -525,6 +525,10 @@ func isHLSFile(fname string) bool {
 	return fname[ext:] == ".m3u8"
 }
 
+func isEncryptionEnabled(params api.UploadTaskParams) bool {
+	return params.Encryption.EncryptedKey != ""
+}
+
 func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.OutputLocation, error) {
 	playbackId := tctx.OutputAsset.PlaybackID
 	outURL := tctx.OutputOSObj.URL
@@ -534,7 +538,7 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
 	} else {
 		mp4 = OUTPUT_ONLY_SHORT
 	}
-	outputNames, outputLocations, err := outputLocations(outURL, OUTPUT_ENABLED, playbackId, mp4, playbackId, true)
+	outputNames, outputLocations, err := outputLocations(outURL, OUTPUT_ENABLED, playbackId, mp4, playbackId, !isEncryptionEnabled(*tctx.Task.Params.Upload))
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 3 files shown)


task/upload.go
```diff
@@ -389,9 +387,6 @@ func processCatalystCallback(tctx *TaskContext, callback *clients.CatalystCallba
 			return nil, fmt.Errorf("unknown output name=%q for output=%+v", outName, output)
 		}
 	}
-	if FlagCatalystCopiesSourceFile && videoFilePath == "" {
-		return nil, fmt.Errorf("no video file path found in catalyst output")
-	}
```
victorges, line 394: Do you mean to only return the `FlagCatalystCopiesSourceFile` check? This was meant to be a double check that Catalyst did actually return the file path for the copied file, if we expect that from it.

task/upload.go
```diff
@@ -528,23 +522,6 @@ func isHLSFile(fname string) bool {
 	return fname[ext:] == ".m3u8"
 }
 
-func removeCredentials(metadata *clients.CatalystCallback) *clients.CatalystCallback {
```
victorges, line 531: Was this only used when saving that `video.json` file?

task/upload.go
```diff
@@ -453,13 +448,12 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
 		return tctx.Progress.TrackReader(rawSourceFile, size, endProgress), nil
 	}
 
+	fullPath := videoFileName(playbackID)
+	assetSpec.Files = append(assetSpec.Files, api.AssetFile{
+		Type: "source_file",
+		Path: toAssetRelativePath(playbackID, fullPath),
+	})
```
victorges, line 455: Could you do this outside this `comlpetementCatalystPipeline` function? The idea for this function, at least originally, was to keep only the features not implemented yet by Catalyst that we needed to complement later. Exactly to facilitate the changes you are making now, knowing exactly the bits that are/were missing from Catalyst. 

So I think it'd make a lot more sense for this logic to be under `processCatalystCallback` instead, not here, since it's not complementing the missing features but only processing the catalyst callback. After the changes you are doing, this function should be completely removed.

WDYT?

task/upload.go
```diff
@@ -576,7 +553,7 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
 	return outputNames, outputLocations, nil
 }
 
-func outputLocations(outURL, hls, hlsRelPath, mp4, mp4RelPath string) ([]OutputName, []clients.OutputLocation, error) {
+func outputLocations(outURL, hls, hlsRelPath, mp4, mp4RelPath string, sourceCopy bool) ([]OutputName, []clients.OutputLocation, error) {
```
victorges, line 556: At some point we might want an options struct arg for this function hahah

task/upload.go
victorges: After reading the rest of the code, it seems like there is no check at all that the source file was actually generated, we just trust that it was because it was asked. Would be good to either have a confirmation from Catalyst (like an explicit output on the callback) or to check here that the file is there, instead of just returning an untested path. WDYT?

victorges (approved):
LGTM anyway, but made some comments regarding the "loosiness" of the current interface and the location where some things are being done (complement vs process).

All comments are optional, feel free to merge as is.

task/upload.go
mjh1: Yep 👍 no more usages of it now

task/upload.go
mjh1: There is a different check in a previous PR here: https://github.com/livepeer/task-runner/pull/204/files#diff-d454bfad6f152c810569d9a69d364cc443c95f99e3e4171b10c088a851c8d3e3R419-R421 so if the file exists from catalyst we use it, otherwise we make the copy here in task-runner. My plan was to have it like this initially to be backwards compatible then tidy everything up afterwards removing the `complementCatalystPipeline` function.

PR #209, opened 2023-08-21
FMP4 support

mjh1:
Allow for FMP4 outputs in the transcode API


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/golang/glog v1.1.1
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/livepeer/catalyst-api v0.1.2-0.20230612153934-449655e3778a
-	github.com/livepeer/go-api-client v0.4.8
+	github.com/livepeer/go-api-client v0.4.9-0.20230818122953-366df6f5d816
```
leszko, line 10: I think you can update it to the `0.4.9` version before merging.

task/upload.go
```diff
@@ -538,7 +546,7 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
 	} else {
 		mp4 = OUTPUT_ONLY_SHORT
 	}
-	outputNames, outputLocations, err := outputLocations(outURL, OUTPUT_ENABLED, playbackId, mp4, playbackId, !isEncryptionEnabled(*tctx.Task.Params.Upload))
+	outputNames, outputLocations, err := outputLocations(outURL, OUTPUT_ENABLED, playbackId, mp4, playbackId, "", "", !isEncryptionEnabled(*tctx.Task.Params.Upload))
```
leszko, line 549: Why do we pass empty strings here? Is it because we don't support fMP4 in Transcode API?

mjh1: The opposite but yes, this is for the normal upload task, we only support fmp4 for transcode API at the moment.

leszko: 👍

leszko (approved):

task/upload.go
```diff
@@ -583,6 +600,17 @@ func outputLocations(outURL, hls, hlsRelPath, mp4, mp4RelPath string, sourceCopy
 				},
 			},
 		}
+	if fmp4 == OUTPUT_ENABLED {
+		names, locations =
+			append(names, OutputNameEmpty),
+			append(locations, clients.OutputLocation{
+				Type: "object_store",
+				URL:  url.JoinPath(videoFileName(fmp4RelPath)).String(),
```
emranemran, line 608: @mjh1 this is adding a /video/fmp4 to the paths which we don't want. We just want the /fmp4 appended to whatever path the user provides.

PR #212, opened 2023-08-29
Remove probing - catalyst adds all the appropriate fields now

mjh1:


victorges (approved):
Niiiiice

task/upload.go
```diff
@@ -480,19 +478,6 @@ func complementCatalystPipeline(tctx *TaskContext, assetSpec api.AssetSpec, call
 		}
 	}
 
-	if !FlagCatalystProbesFile {
```
victorges, line 483: I'm almost sure there is one huge optimization we can do now: if catalyst copied the source file (`catalystCopiedSource`) and there is no IPFS storage (L481), we don't even need to download the source file anymore! Can probably do an early return around L447.

Haven't double checked alllll the logic that still remains here tho, so do make sure, but it seems overall possible 👀

mjh1:
@victorges does the final logic look ok to you?

victorges:
@mjh1 Sorry didnt see the comment til now. It does look good!

PR #213, opened 2023-09-01
Don't reject audio-only VOD assets

thomshutt:
This is now supported in Catalyst API with https://github.com/livepeer/catalyst-api/pull/843

### DIFF
--- a/task/probe.go
+++ b/task/probe.go
@@ -98,9 +98,6 @@ func toAssetSpec(filename string, strict bool, probeData *ffprobe.ProbeData, siz
 		hasAudio = hasAudio || track.Type == "audio"
 		spec.VideoSpec.Tracks = append(spec.VideoSpec.Tracks, track)
 	}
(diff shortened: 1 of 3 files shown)


mjh1 (approved):

mjh1:
This probe call is not longer used by the upload task, however I did notice it's still in use by 'transcode' (not transcode API) and 'import' tasks. @victorges could you remind me are these tasks in use anymore? If not would it be safe to delete them?

victorges:
@mjh1 Those are the legacy flows for VOD. Import is equivalent to the Upload today, while transcode was a "mutation" of existing assets like to ask for a different profile for the master rendition or something like that. I believe it is not used anywhere, we don't document it nor expose on SDKs at least, and I guess they might even be completely broken by now. You can make sure by looking for tasks of type `import` or `transcode` on the DB, but I think it should be mostly safe to delete yeah!

PR #215, opened 2023-09-18
Remove unused tasks Import and Transcode

mjh1:
Import was previously used for recordings v1
Transcode was another legacy flow, for creating a "mutation" of existing assets e.g. to ask for a different profile for the master rendition


task/transcode.go
```diff
@@ -1,342 +1,13 @@
...
-)
-
-const (
-	segLen               = 2 * time.Second
-	maxFileSizeForMemory = 50_000_000
 )
 
 func init() {
 	format.RegisterAll()
 	rand.Seed(time.Now().UnixNano())
```
mjh1, line 12: @victorges Any idea why this was here? It's a deprecated function anyway

victorges: I think this was probably copied from the record-tester code that this is based of: https://github.com/livepeer/stream-tester/blob/9b92a073ee2dd84323f0c8c8686458c3319b33c0/cmd/recordtester/recordtester.go#L50C34-L50C34

It was included in the first PR for this https://github.com/livepeer/task-runner/pull/7/files#diff-d09eb25053ed0ed141e44c9262f4138cb013dd9b1a386428d27f9366c0e44487R32 without much explanation nor use of `rand` package anywhere else, so I think it might've been a miscopy

task/transcode.go
```diff
@@ -1,342 +1,13 @@
...
-	"github.com/livepeer/stream-tester/segmenter"
-)
-
-const (
-	segLen               = 2 * time.Second
-	maxFileSizeForMemory = 50_000_000
 )
 
 func init() {
 	format.RegisterAll()
```
victorges, line 11: Are we using joy4 anywhere else? If not we should be able to remove that dependency as well!!

task/transcode.go
victorges, line 1: Any reason to still keep this file? If the only bit left is they init(), might as well move that to somewhere it is relevant (if there's anywhere using joy4)

task/upload.go
```diff
@@ -1,36 +1,56 @@
...
+	segLen               = 2 * time.Second
+	maxFileSizeForMemory = 50_000_000
 )
 
 var (
 	// Feature flag whether to use Catalyst's IPFS support or not.
 	FlagCatalystSupportsIPFS = false
 )
 
+type ImportTaskConfig struct {
```
victorges, line 49: WDYT of renaming all these from `Import` to `Upload` now? :D

task/upload.go
```diff
@@ -626,3 +646,192 @@ func catalystTaskAttemptID(task *api.Task) string {
 	// something more sophisticated in the future.
 	return strconv.Itoa(task.Status.Retries)
 }
+
+func getFile(ctx context.Context, osSess drivers.OSSession, cfg ImportTaskConfig, params api.UploadTaskParams, vodDecryptPrivateKey string) (name string, size uint64, content io.ReadCloser, err error) {
```
victorges, line 650: I kinda liked having these lower level helpers separate not to clutter the main task logic (which is already huge). WDYT of keeping these in a helper file like `files.go` or sth like that?

(btw other than this comment, I'm just assuming the content is exactly the same as transcode.go funcs)

victorges (approved):
Great cleanup!

BTW, one thing you also have to do is remove the creation of such deleted tasks from the Studio API. Both for cleaning up that code as well but also to avoid weird errors in case these APIs eventually get called

task/transcode.go
mjh1: yeah i can't see it being used, removed

task/probe.go
```diff
@@ -1,266 +0,0 @@
...
-		add("duration", stream.Duration)
-		if stream.CodecType == "video" {
-			add("pixelFormat", stream.PixFmt)
-			width, height = stream.Width, stream.Height
-		}
-		if startTime, err := strconv.ParseFloat(stream.StartTime, 64); err == nil {
-			maxStartTime = math.Max(maxStartTime, startTime)
-		}
-	}
-	glog.Infof("Probed video file assetId=%s filename=%q format=%q width=%d height=%d bitrate=%s startTime=%v maxStreamStartTime=%v %s",
```
pwilczynskiclearcode, line 254: @mjh1 I guess [this alert](https://eu-metrics-monitoring.livepeer.live/grafana/alerting/LNUr4PKVk/edit?returnTo=%2Falerting%2Flist%3Fsearch%3D) is not needed since this line got removed?

mjh1: Yeah i'd remove, must not be relevant anymore, I'm not aware we're checking for that condition on the alert anymore

PR #217, opened 2023-09-21
clip: added clip task handling

gioelecerati:


emranemran (approved):

task/upload.go
```diff
@@ -47,6 +47,7 @@ var (
 	OutputNameOSSourceMP4   = OutputName("source_mp4")
 	OutputNameOSPlaylistHLS = OutputName("playlist_hls")
 	OutputNameIPFSSourceMP4 = OutputName("ipfs_source_mp4")
+	OutputNameClipSource    = OutputName("clip_source")
```
victorges, line 50: Are clip outputs different from regular MP4/HLS? I thought it was everything the same, only changing the input. We could re-use a lot of logic if they were the same, like on the API to process the task output.

task/upload.go
```diff
@@ -72,27 +74,57 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
...
 
-		if params != nil && params.Encryption.EncryptedKey != "" {
-			encryption = &clients.EncryptionPayload{
-				EncryptedKey: params.Encryption.EncryptedKey,
+		var encryption *clients.EncryptionPayload
+		var clipStrategy *video.ClipStrategy
+
+		if tctx.Task.Params.Clip != nil {
+			clipParams := tctx.Task.Params.Clip
+			if clipParams != nil {
```
victorges, line 83: duplicate check
```suggestion
		if clipParams := tctx.Task.Params.Clip; clipParams != nil {
```

task/upload.go
```diff
@@ -72,27 +74,57 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
...
+				clipStrategy = &video.ClipStrategy{
+					StartTime:  clipParams.ClipStrategy.StartTime,
+					EndTime:    clipParams.ClipStrategy.EndTime,
+					PlaybackID: clipParams.ClipStrategy.PlaybackId,
+				}
```
victorges, line 88: Isn't this the exact same type+object? Why do we need to recreate it? Also on `TaskClip`

task/upload.go
```diff
@@ -72,27 +74,57 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
...
-				Url:                   inUrl,
-				CallbackUrl:           tctx.catalyst.CatalystHookURL(tctx.Task.ID, "finalize", catalystTaskAttemptID(tctx.Task)),
-				OutputLocations:       outputLocations,
-				PipelineStrategy:      p.catalystPipelineStrategy,
-				Profiles:              p.profiles,
-				TargetSegmentSizeSecs: p.targetSegmentSizeSecs,
-				Encryption:            encryption,
+
+		var req clients.UploadVODRequest
+		req = clients.UploadVODRequest{
```
victorges, line 100:
```suggestion
		req := clients.UploadVODRequest{
```

task/upload.go
```diff
@@ -72,27 +74,57 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
...
+				EndTime:    clipStrategy.EndTime,
+				PlaybackID: clipStrategy.PlaybackID,
+			}
+		} else {
+			// TODO : this is just temporary for staging - to remove
+			req.ClipStrategy = clients.ClipStrategy{
+				StartTime:  0,
+				EndTime:    0,
+				PlaybackID: "",
 			}
```
victorges, line 123: This is not necessary. Go already initializes objects with the 0 value. I think it was a noop even in staging.

task/upload.go
```diff
@@ -72,27 +74,57 @@ func handleUploadVOD(p handleUploadVODParams) (*TaskHandlerOutput, error) {
...
-				ExternalID:            tctx.Task.ID,
-				Url:                   inUrl,
-				CallbackUrl:           tctx.catalyst.CatalystHookURL(tctx.Task.ID, "finalize", catalystTaskAttemptID(tctx.Task)),
-				OutputLocations:       outputLocations,
-				PipelineStrategy:      p.catalystPipelineStrategy,
-				Profiles:              p.profiles,
-				TargetSegmentSizeSecs: p.targetSegmentSizeSecs,
-				Encryption:            encryption,
+
+		var req clients.UploadVODRequest
```
victorges, line 99: Maybe at some point we just want to move this `clients.UploadVODRequest` to the `handleUploadVODParams` and let callers create their catalyst requests. Otherwise we have this "generic" "handle upload" function that is not really generic, just has a bunch of conditionals to implement all the possible task types this might run. Might as well not have the abstraction if it grows like that.

task/upload.go
```diff
@@ -264,6 +323,29 @@ func toTranscodeFileTaskOutput(callback *clients.CatalystCallback) (data.Transco
...
+	}
+	// we expect only one output
+	o := callback.Outputs[0]
+
+	bu, p, err := parseUrlToBaseAndPath(o.Manifest)
+	if err != nil {
+		return res, err
+	}
+	res.BaseUrl = bu
+	if len(o.Videos) > 0 {
```
victorges, line 342: So clips may or may not have HLS, is that it? Does that mean assets can also have only an Mp4 now but not HLS? Or is it the other way around?

IOW what happens if there is no videos here? We return a base URL that points to a folder but there are no files there? Shouldn't that be an error?

task/upload.go
```diff
@@ -573,14 +655,34 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
...
 }
 
+func clipTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.OutputLocation, error) {
+	playbackId := tctx.OutputAsset.PlaybackID
+	sourcePlaybackId := tctx.Task.Params.Clip.ClipStrategy.PlaybackId
+	sessionId := tctx.Task.Params.Clip.InputSessionID
+	outURL := tctx.OutputOSObj.URL
+	var mp4 string
+
+	clipRelPath := sourcePlaybackId + "/" + sessionId + "/clip_" + playbackId
```
victorges, line 665:
```suggestion
	clipRelPath := path.Join(sourcePlaybackId, sessionId, "clip_" + playbackId)
```

victorges (approved):

task/upload.go
emranemran: We just needed this temporarily because I was inadvertently setting StartTime to be a required value on catalyst-api. We can remove this now.

task/upload.go
victorges: Just to clarify, the step 3 is provided like a "source playback" workflow right? Or would the job complete from the `task-runner` side but stay running on Catalyst?

task/upload.go
victorges: I only meant that this assignment here is a noop. It will not change what is the serialization of the `ClipStrategy` field since it doesn't change the value of any of its fields (and it's stored in `req` by value, not by pointer)

What would really make a difference is the `omitempty` flag on the json tag (but even then, this assignment is noop). A playground: https://go.dev/play/p/4z6mTYrk59_Y

task/upload.go
gioelecerati: We changed everything with output locations and now the clip task output is going to go with the same output locations of VOD, + the clip source

task/upload.go
gioelecerati: This is also outdated because of last night changes

task/upload.go
victorges: Up. We have double the complexity and indentation here.

task/upload.go
```diff
@@ -57,6 +58,7 @@ type handleUploadVODParams struct {
 	profiles                 []api.Profile
 	targetSegmentSizeSecs    int64
 	catalystPipelineStrategy pipeline.Strategy
+	clipStrategy             video.ClipStrategy
```
victorges, line 61: Doesn't seem to be a need for using the `catalyst-api` type here, since we are redefining it on `clients` and have to convert to that one in the end. So I'd say this should be just `*clients.ClipStrategy` (a pointer) and you can skip a couple of object recreations as well.

task/upload.go
```diff
@@ -216,12 +247,39 @@ func TaskTranscodeFile(tctx *TaskContext) (*TaskHandlerOutput, error) {
...
+			return outputLocations, err
+		},
+		finalize: func(callback *clients.CatalystCallback) (*TaskHandlerOutput, error) {
+			tctx.Progress.Set(0.9)
+			taskOutput, err := processCatalystCallback(tctx, callback)
+			if err != nil {
+				return nil, fmt.Errorf("error processing catalyst callback: %w", err)
+			}
+			return &TaskHandlerOutput{
+				TaskOutput: &data.TaskOutput{Upload: taskOutput},
```
victorges, line 271: This should be `Clip` or API will break processing this result.
```suggestion
				TaskOutput: &data.TaskOutput{Clip: taskOutput},
```

task/upload.go
```diff
@@ -573,6 +631,38 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("error parsing clip URL: %w", err)
+	}
+
+	outputNames, outputLocations, err := outputLocations(outURL, OUTPUT_ENABLED, playbackId, OUTPUT_ENABLED, playbackId, "", "", false)
+
+	if err != nil {
+		return nil, nil, err
+	}
+	clipOutputLocationUrl := clipURL.JoinPath("/clip_" + playbackId).String()
```
victorges, line 651: Won't this make the clips ephemeral? Why aren't clips stored in a similar pattern as any other asset? i.e. on the VOD object store & with a path like `/hls/<playback_id>/...` (the asset playback ID, not the stream)

Being out of context, it seems like clips are just a special kind of upload with start/end timestamps. The outputs should be just the same tho and go to the same place as regular VODs. We also don't need to make the files "related" on the object store, like storing clips inside the stream folder, just need them to be related in the API (which they already are, through `source.sessionId`). I'm probably missing something though.

victorges (approved):
LGTM

task/upload.go
gioelecerati: Yes - this is the temporary location for the processing clip that I was referring on the studio PR - the actual final HLS and MP4 coming out of the vod job will be in the normal path for assets

task/upload.go
tritone11: Exactly like a source playback for assets!

gioelecerati: Exactly like a source playback for assets!

PR #223, opened 2023-10-20
Support for VOD thumbnails

mjh1:
Plus refactored the `outputLocations` function a little to reduce the repetition and growing param list.

### DIFF
--- a/task/upload.go
+++ b/task/upload.go
@@ -219,10 +219,11 @@ func TaskTranscodeFile(tctx *TaskContext) (*TaskHandlerOutput, error) {
 		getOutputLocations: func() ([]clients.OutputLocation, error) {
 			_, outputLocation, err := outputLocations(
 				params.Storage.URL,
-				out(isEnabled(params.Outputs.HLS.Path), params.Outputs.HLS.Path),
-				out(isEnabled(params.Outputs.MP4.Path), params.Outputs.MP4.Path),
-				out(isEnabled(params.Outputs.FMP4.Path), params.Outputs.FMP4.Path),
-				output{},
+				outputs{
+					hls:  out(isEnabled(params.Outputs.HLS.Path), params.Outputs.HLS.Path),
+					mp4:  out(isEnabled(params.Outputs.MP4.Path), params.Outputs.MP4.Path),
+					fmp4: out(isEnabled(params.Outputs.FMP4.Path), params.Outputs.FMP4.Path),
+				},
 				false,
 			)
 			return outputLocation, err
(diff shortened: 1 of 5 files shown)


task/upload.go
```diff
@@ -677,7 +675,11 @@ func uploadTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.Outpu
 	} else {
 		mp4 = OUTPUT_ONLY_SHORT
 	}
-	outputNames, outputLocations, err := outputLocations(outURL, OUTPUT_ENABLED, playbackId, mp4, playbackId, "", "", !isEncryptionEnabled(*tctx.Task.Params.Upload))
+	thumbsEnabled := ""
+	if tctx.Task.Params.Upload.Thumbnails {
+		thumbsEnabled = OUTPUT_ENABLED
+	}
+	outputNames, outputLocations, err := outputLocations(outURL, out(OUTPUT_ENABLED, playbackId), out(mp4, playbackId), output{}, out(thumbsEnabled, playbackId), !isEncryptionEnabled(*tctx.Task.Params.Upload))
```
victorges, line 682: nit: Mind break this down 1 arg per line so its more readable? Like on `TaskTranscodeFile`

task/upload.go
```diff
@@ -738,14 +740,26 @@ func clipTaskOutputLocations(tctx *TaskContext) ([]OutputName, []clients.OutputL
...
 func outputLocations(
-	outURL,
+	outURL string,
 	hls,
-	hlsRelPath,
 	mp4,
-	mp4RelPath,
 	fmp4,
-	fmp4RelPath string,
+	thumbnails output,
```
victorges, line 762: WDYT of using `*output` so we can use `nil` as the "not configured" value, instead of an empty struct?

task/upload.go
```diff
@@ -755,46 +769,35 @@ func outputLocations(
...
-				URL:  url.JoinPath(mp4RelPath).String(),
-				Outputs: &clients.OutputsRequest{
-					MP4: mp4,
-				},
-			},
+			outputLocation(url, hls.path, &clients.OutputsRequest{HLS: hls.enabled}),
+			outputLocation(url, mp4.path, &clients.OutputsRequest{MP4: mp4.enabled}),
 		}
-	if fmp4 == OUTPUT_ENABLED {
+	if fmp4.enabled == OUTPUT_ENABLED {
```
victorges, line 775: Is this `enabled` field always `OUTPUT_ENABLED` or an empty string (or sth that means false)? Could be worth making it a boolean in the `output` struct to simplify, if thats the case

victorges (approved):
LGTM, nice refact 🙏 
Comments are all nits/style, feel free to merge as is.

task/upload.go
mjh1: An enum would be nice but then we'd still have to convert to the string for catalyst before sending the request.

task/upload.go
mjh1: I wanted to avoid pointers so that nil checks aren't necessary etc but I've introduced a "holder" struct which means no need to explicitly type out the empty struct def. Hope that looks good to you

victorges: Looks great!

PR #226, opened 2023-11-20
Enable thumbs for all studio vod jobs

mjh1:


emranemran:
LGTM, main concern is the ffmpeg calls on every segment which will increase our pod loads.

PR #233, opened 2024-03-08
runner: cron job for asset deletion

gioelecerati:


task/runner.go
```diff
@@ -545,6 +546,97 @@ func (r *runner) Shutdown(ctx context.Context) error {
...
+				if err != nil {
+					glog.Errorf("Error listing files for asset %v: %v", asset.ID, err)
+					continue
+				}
+
+				for pi != nil {
+					for _, file := range pi.Files() {
+						err := assetOS.DeleteFile(ctx, file.Name)
+						if err != nil {
+							glog.Errorf("Error deleting file %v: %v", file.Name, err)
```
victorges, line 604: Reminder to not tag the asset as deleted if there are any errors deleting file

task/runner.go
```diff
@@ -545,6 +546,97 @@ func (r *runner) Shutdown(ctx context.Context) error {
...
+						if err != nil {
+							glog.Errorf("Failed to load next page of files for asset %v: %v", asset.ID, err)
+							break
+						}
+					} else {
+						break // No more pages
+					}
+				}
+
+				glog.Infof("Deleted %v files from asset=%v", totalDeleted, asset.ID)
```
victorges, line 621: Gotta tag asset as deleted (and probably unpin before considering it done)

task/runner.go
```diff
@@ -545,6 +546,97 @@ func (r *runner) Shutdown(ctx context.Context) error {
 	return r.amqp.Shutdown(ctx)
 }
 
+func (r *runner) UnpinFromIpfs(ctx context.Context, cid string, filter string) error {
```
victorges, line 549: Can we move these new funcs to a separate file/struct? I think this runner is already doing a lot so we can keep business logic separate

task/runner.go
victorges: As a general technique I like when creating background jobs like this, you could create an internal version of the function that returns an error like other functions normally do, then have a "background runner" function that only handles calling that function and logging whatever error it returns. It generally makes the code less repetitive with all the logs and less error-prone (e.g. if you do an early return on sth that should log). 

No need to implement this tho, sending as a suggestion in case you find it useful when you refactor it to a separate file/out of `runner`

victorges (approved):
LGTM, all comments but some minor ones (tagging asset as deleted) are optional

